PR #8, opened 2021-08-24
Add Kitty themes

mrjones2014:
Adds a theme for Kitty terminal emulator, using the catppuccino theme colors.

PR #9, opened 2021-08-24
Add Alacritty themes

knpwrs:
This PR adds some themes for Alacritty!

Some questions:

1. I noticed some extended colors in #8, should those be replicated here?
2. I noticed that the actual background of the text editor area in neovim is darker than the `bg` color listed in `catppuccino.lua`. In my personal Alacritty configuration I used `#121212` as the background so it would be darker to match neovim. Do you have a preference?
3. It looks like in the lua files in the middle of all the bright colors the bright red color is called `red_bg` -- is this intentional? The other bright colors surrounding it are all called `_br`.

### DIFF
--- a/extra/alacritty-catpuccino.yml
+++ b/extra/alacritty-catpuccino.yml
@@ -1,0 +1,35 @@
+# Catpuccino
+colors:
+  # Default colors
+  primary:
+    background: '0x0e171c'
+    foreground: '0xabb2bf'
+
+  # Colors the cursor will use if `custom_cursor_colors` is true
+  cursor:
+    text: '0x0e171c'
+    cursor: '0xabb2bf'
+
+  # Normal colors
+  normal:
+    black:   '0x393b44'
+    red:     '0xc94f6d'
+    green:   '0x97c374'
+    yellow:  '0xdbc074'
+    blue:    '0x61afef'
+    magenta: '0xc678dd'
+    cyan:    '0x63cdcf'
+    white:   '0xdfdfe0'
+
+  # Bright colors
+  bright:
+    black:   '0x7f8c98'
+    red:     '0xe06c75'
+    green:   '0x58cd8b'
+    yellow:  '0xffe37e'
+    blue:    '0x84cee4'
+    magenta: '0xb8a1e3'
+    cyan:    '0x59f0ff'
+    white:   '0xfdebc3'
+
+  # indexed_colors: []
--- a/extra/alacritty-light-melya.yml
+++ b/extra/alacritty-light-melya.yml
@@ -1,0 +1,35 @@
+# Catppuccino Light Melya
+colors:
+  # Default colors
+  primary:
+    background: '0xfbfbfb'
+    foreground: '0x0e171c'
+
+  # Colors the cursor will use if `custom_cursor_colors` is true
+  cursor:
+    text: '0x0e171c'
+    cursor: '0xfbfbfb'
+
+  # Normal colors
+  normal:
+    black:   '0x393b44'
+    red:     '0xb0304e'
+    green:   '0x76ab49'
+    yellow:  '0xffce1f'
+    blue:    '0x157c8e'
+    magenta: '0xa414cc'
+    cyan:    '0x63cdcf'
+    white:   '0xdfdfe0'
+
+  # Bright colors
+  bright:
+    black:   '0x7f8c98'
+    red:     '0xd84652'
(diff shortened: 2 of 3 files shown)


mrjones2014:
@knpwrs the bg issue probably has to do with Alacritty or your config. It matches for me using kitty, tmux, and neovim.

Pocco81:
> I noticed some extended colors in Add Kitty themes #8, should those be replicated here?

Not necessarily

> It looks like in the lua files in the middle of all the bright colors the bright red color is called red_bg -- is this intentional? The other bright colors surrounding it are all called _br.

Nope, it's a error. Excuse me for that, I'll fix it ASAP :)

Pocco81:
> I noticed that the actual background of the text editor area in neovim is darker than the bg color listed in catppuccino.lua

Hmm that's weird, the color must match. Here is an example with Neon Latte. Both the color of the current bg and `bg` in Neon Latte's settings match:

image

I also agree with @mrjones2014.

mrjones2014:
@knpwrs are you using tmux? are you using term or gui colors in nvim? what is the value of `$TERM` in your shell?

Running `:checkhealth` in nvim might help diagnose the issue.

knpwrs:
* I am using tmux (3.2)
* `:set termguicolors?` returns `termguicolors` and not `notermguicolors`
* `echo $TERM` returns `screen-256color`, outside of tmux it returns `xterm-256color`
* I did have `set -g default-terminal "screen-256color"` in my `tmux.conf`. Changing this to `set -g default-terminal "xterm-256color"` does not fix the issue inside of tmux
* It looks like running `nvim` outside of tmux I get the proper background color ❗ 
* I didn't see anything relevant in `:checkhealth`
* Here is my color picker on macOS, for reference:
  image
* Here is the color picker when I run `nvim` outside of tmux:
  image

I think I'm starting to understand why vim themes have always given me so much trouble...

So it's something tmux related? I'm admittedly weak in this area, I'm a heavy terminal / tmux / vim user but I don't know much about the internals of how theming works. I just upgraded tmux to 3.2.9 but no dice.

This is the current state of my `tmux.conf` file, except that locally I have line 58 as `set -g default-terminal "xterm-256color"` instead of `set -g default-terminal "screen-256color"` (since I was just debugging all this): https://github.com/knpwrs/dotfiles/blob/c55697211270e0b9dab8b2caf6a43cb2208075cc/home/tmux.conf

mrjones2014:
@knpwrs try, in `tmux.conf`, `set-option -g default-terminal "screen-256color"` (`set-option` as opposed to `set`), then exit any tmux sessions you have open, restart the tmux server, and try that.

knpwrs:
@mrjones2014 no dice, both `set-option -g default-terminal "screen-256color"` and `set-option -g default-terminal "xterm-256color"` don't fix the colors from inside tmux. I'm making sure to run `tmux kill-server` after exiting tmux to make sure I'm starting fresh.

mrjones2014:
Hm, @knpwrs try including _both_ of these lines, just looked at my config and I have both (I also have these additional 2 lines but those are for under curls, don't think they affect color, but maybe try them anyway?):

```
# Make Neovim 256 color work properly
# Make undercurls work properly in Neovim
# Undercurl
set-option -g default-terminal "screen-256color"
set -g default-terminal "screen-256color"
set -as terminal-overrides ',*:Smulx=\E[4::%p1%dm'  # undercurl support
set -as terminal-overrides ',*:Setulc=\E[58::2::%p1%{65536}%/%d::%p1%{256}%/%{255}%&%d::%p1%{255}%&%d%;m'  # underscore colours - needs tmux-3.0
```

knpwrs:
Still no dice :(

image

I personally like the darker color better, but I really would like to figure out what is going on here.

However, this just in: Kitty has the right background color with tmux. Here are Kitty and Alacritty side by side attached to the same tmux session:

image

So this is a combination tmux/alacritty thing. I just upgraded Alacritty from 0.8.0 to 0.9.0 and still no dice.

knpwrs:
Huzzah! I got it! With the following two lines in `tmux.conf` I get the right colors in Alacritty:

```
set -g default-terminal "screen-256color"
set-option -ga terminal-overrides ",screen-256color:Tc"
```

knpwrs:
With that out of the way I think this PR should be clear, yeah? @Pocco81 I know you said the indexed colors don't necessarily need to be ported over but would it make sense to do for for consistency anyway?

PR #27, opened 2021-09-03
Define hop colors

SmiteshP:
Defined colors for hop.nvim

Default hop colors  (with dark_catppuccino) :
Screenshot from 2021-09-03 09-10-12

After this commit
dark_catppuccino:
Screenshot from 2021-09-03 09-10-30

soft_manilo
Screenshot from 2021-09-03 09-10-51

neon_latte
Screenshot from 2021-09-03 09-11-04

light_melya
Screenshot from 2021-09-03 09-11-30

In my opinion these colors look good in the dark themes, but not so good in light_melya.  Is it possible to define different colors only for light_melya?

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ integrations = {
 	markdown = false,
 	lightspeed = false,
 	ts_rainbow = false,
+	hop = false,
 }
 ```
 
@@ -242,6 +243,7 @@ catppuccino.setup(
 			markdown = false,
 			lightspeed = false,
 			ts_rainbow = false,
+			hop = false,
 		}
 	}
 )
@@ -302,6 +304,7 @@ catppuccino.setup(
 			markdown = false,
 			lightspeed = false,
 			ts_rainbow = false,
+			hop = false,
 		}
 	}
 )
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@
     -   [Lightspeed](https://github.com/ggandor/lightspeed.nvim)
     -   [Nvim-ts-Rainbow](https://github.com/p00f/nvim-ts-rainbow)
     -   [Sneak](https://github.com/justinmk/vim-sneak)
+    -   [Hop](https://github.com/phaazon/hop.nvim)
     -   [Neogit](https://github.com/TimUntersberger/neogit)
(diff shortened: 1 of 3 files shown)


Pocco81:
This looks sick man! thanks for the PR :)

Pocco81:
> In my opinion these colors look good in the dark themes, but not so good in light_melya

I personally feel like it's okay for the light theme? I mean, they are very distinguishable, aren't they?

image

>  Is it possible to define different colors only for light_melya?

Yes but in case you wanted to do this the best approach would be to determine whether the background is light or dark. To do that you can use Catppuccino's `colors` utils like so:

```lua
local colors_util = require("catppuccino.utils.colors")
if colors_util.assert_brightness(cpt.bg) then
	-- stuff
end
```

And then you'd use this to create a variable to change to if the background is bright.

In your opinion @SmiteshP, do you think it's hard to see?

SmiteshP:
> In your opinion @SmiteshP, do you think the colors over the bright background are hard to see?

Yep, thats what I felt. I think that adding darker background for the hop highlights would make it better, something like in incsearch.

Pocco81:
Yep I forgot... but anyways, a5a3d62f73eb2fe98b6353b62f26d51109c5e289 should've improved this :+1:

SmiteshP:
Welp, I actually don't like this change tbh 😅 
Underlining the HopNextKey isn't necessary. 
And HopNextKey2 should look subtle compared to HopNextKey1, thats why I had made it use a darker shade of blue and without any special style. Right now the 2 letter jumps seem like a lot of visual clutter.

That's just my opinion, if you like it this way its fine too 🤷🏽‍♂️

Pocco81:
> Underlining the HopNextKey isn't necessary.

I saw that other plugins like hop use the underline for the first char thing. 

> Right now the 2 letter jumps seem like a lot of visual clutter.

But then they'd be hard to see in light backgrounds, wouldn't they?

Unlike you, I don't use Hop.nvim, so feel free to PR better colors :)

SmiteshP:
> Unlike you, I don't use Hop.nvim, so feel free to PR better colors :)

Sure, I will try to make another PR with better colors for light_melya, while keeping the orange and blue combo for darker themes 😄

PR #33, opened 2021-09-10
Fix lightline not finding colorscheme

gotgenes:
This fixes the issue where lightline complains it can't find the catppuccino colorscheme.

The colorscheme must be stored in a global variable that lightline expects.


Pocco81:
Thanks for the PR mate.

gotgenes:
@Pocco81 Thanks for providing Catppuccino and for accepting my PR!

Issue #36, opened 2021-09-24
add TSInclude highlight group as part of keywords


LuisxSullivaN:
Hi, when I set the keywords in my config to be italic I would like the "import" and "from" keyword (in case of JavaScript) to be italic as well:
![image](https://user-images.githubusercontent.com/29529436/134626053-c74dbc34-d150-4110-9f11-177c239a4b6c.png)

I can achieve this by overriding the highlight `TSInclude` in the config but I think this should be a default since those are keywords from the language and that highlight might apply to other languages as well.

Pocco81:
Good idea! I'll add it ASAP :)

LuisxSullivaN:
Hey! I have another use case:
 I would like all the keywords to be bold and italic, while I would like the function names to be just italic. For that I set my config like this:
![image](https://user-images.githubusercontent.com/29529436/134708827-30277f81-5d50-4f13-982c-10b70739d3de.png)
But I'm getting this result:
![image](https://user-images.githubusercontent.com/29529436/134708980-453e5037-d8b8-4bae-8383-05ca11253ca7.png)

As you can see, functions are italic as expected but the keyword 'function' is not bold, just italic. Making the distinction between `TSFunction` and `TSKeywordFunction` should work.

LuisxSullivaN:
And also, speaking about keywords, what about conditionals? shouldn't 'if' and 'else' be part of the keywords group? and what about loops? haha I may be getting too abroad with this but I think it's a semantic thing and adds the possibility to have all these keywords around our code with a different and very nice looking.

Pocco81:
Oh those keywords are bright red... But perhaps athor colors might look good...? I was thinking about changing some of the colors actually (keywords and loops mostly), but I'm not sure which one(s) might look appealing.

Which color(s) would you recommend?

> What do you think? I could do a PR if needed.

Feel free to contribute. Help is always much appreciated :)

Pocco81:
`TSInclude`:

![image](https://user-images.githubusercontent.com/58336662/134842659-db148d9d-5f93-420e-927d-7c59bc6f2390.png)

Pocco81:
Okay so now `TSFunction` will follow the styles from `function` in the settings and `TSKewordFunction` will follow the styles from `keywords` in the settings.

PR #39, opened 2021-10-05
feat: style missing keywords

LuisxSullivaN:
Because of semantic concerns, style from config should be applied to all
kinds of keywords separated by different highlight groups by Treesitter


LuisxSullivaN:
Hey! @Pocco81, this is what I was talking about in #36, to apply styling to all keywords. We can start with this and then figure out which colours look nicer, what do you think?

Pocco81:
Ooh ok ok now I understand. Are those all the keywords? Do have a list or something I could see? :)

Pocco81:
Great, I'll merge it :)

PR #40, opened 2021-10-07
Use correct diff from colorscheme

AZanellato:
Hey! I saw that on fugitive Git panel the highlight wasn't working properly. After finding the highlight group I saw that the plugin was setting but checked and it seems that the color schemes were using a different word.

My change was to use `delete` instead of remove and not alter the colorschemes. If you want me to change it on the colorschemes, let me know :)

Love the theme!


Pocco81:
I like the change! Thanks for letting me about this issue and thanks for fixing it :)

PR #48, opened 2021-11-02
Respect transparency by providing a value for cp.none

cwebster2:
With the refactor removing values from the color mapper table,
the missing value on cp.none prevents transparency from being
respected.  This adds a value for that in a core.mapper close
to where it is used and does not pollute th color mapping table.


Pocco81:
Oh, thanks, I forgot to add it :)

PR #69, opened 2021-12-13
Adding to TreeSitter colors

jorgebef:
I have added some missing TS colors and modified TSTagAttribute since it was pulling from TSProperty and for certain filetypes like tsx or jsx, it was providing too much red color and having hard contrast.


Pocco81:
Thanks for the nice PR, I think I was having this issue too 👍

PR #71, opened 2021-12-16
changing color for keywords for TreeSitter to catpuccin.4

jorgebef:
I am requesting this PR in order to update the keywords variable in the treesitter integration to change to catpuccin.4 (pink color).

In commit ff28039 the local variable `keywords` was introduced, unifying all the TSKeyword* highlights which is a great idea, however, I feel that using the purple tone color takes away part of the contrast which makes this theme great.

Reintroducing pink tones I believe maintains the contrast low while still providing a little pop to the colorscheme making it more appealing and also easier to read and differentiate.
Otherwise, it feels quite blue based and loses a bit of warmth.

### DIFF
--- a/lua/catppuccin/core/integrations/treesitter.lua
+++ b/lua/catppuccin/core/integrations/treesitter.lua
@@ -1,7 +1,7 @@
 local M = {}
 
 function M.get(cp)
-	local keywords = cp.catppuccin3
+	local keywords = cp.catppuccin4
 	return {
 		-- These groups are for the neovim tree-sitter highlights.
 		-- As of writing, tree-sitter support is a WIP, group names may change.


Pocco81:
@jorgebef First of all, sorry for replying late, I think I saw that you commented on the commit message but I couldn't really find where you commented in the first place. So, sorry about that :)

Now, I'm not sure if you are aware but in the upcoming release the palette receive a couple of fixes/updates (to fix luminance balence mostly). I've discussed this with other members of the community in the [Discord server](https://discord.gg/r6Mdz5dpFc). Do you mind moving the discussion there for the time being? I feel like it'd be a lot easier and afterwards we can post the resolution here. If not then it's fine. I can share with you the details here :+1:

jorgebef:
Closing as we are discussing further changes to the whole highlighting in Discord and creating new PRs

Pocco81:
For the record, the result of this discussion was 13a53888ac9452b4c811ecfa918b7398037c06ee and #73.

PR #90, opened 2022-01-10
fix(ts-rainbow): respect transparent_background

garcia5:
When setting nvim-ts-rainbow highlight groups, do not set a background highlight if  is


Pocco81:
good idea. Thanks for the PR!

PR #91, opened 2022-01-10
fix(ts-rainbow-background): fix guibg highlight

garcia5:
return "NONE" instead of nil


garcia5:
Checked the last PR incorrectly and forgot to remove my personal highlight group override. Sorry about that!

PR #97, opened 2022-01-21
Add a terminal port of the scheme: kitty

theteachr:
- Creates a dir: `extra`.
- Adds a `README` for the dir.
- Adds `kitty.conf` with catppuccin colors.

### DIFF
--- a/extra/README.md
+++ b/extra/README.md
@@ -1,0 +1,3 @@
+# Catppuccin Extras
+
+Terminal ports of catppuccin.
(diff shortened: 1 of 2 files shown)


Pocco81:
Sorry but this isn't needed. There is a [list with all the available ports](https://github.com/catppuccin/catppuccin#ports-and-more) in the main repo's README, including one for Kitty: https://github.com/catppuccin/kitty

PR #111, opened 2022-03-22
fix: Handle nil winid values in feline integration

otherJL0:
Attempting to use the feline integration resulted in an error message when `winid` was `nil`. This pr contains the following changes:

* Applying stylua formatting
* Added helper function `is_enabled` that ensures winid is always a Lua number
* Replacing blocks setting the `enabled` attribute with calls to the new `is_enabled` helper function.

Let me know if the styling or naming should be tweaked.


Pocco81:
thank you! can confirm the fix works :)

PR #119, opened 2022-04-18
README added requirement notice to feline integration section

mafejzuli:
The feline integration needs the kyazdani42/nvim-web-devicons plugin to function properly,
therefore the reader should be warned about the requirement.


Pocco81:
You are right. Thanks for the PR!

PR #138, opened 2022-05-27
tweaked telescope colors and added custom highlights

BattleCh1cken:
It bothered me that the telescope popup blended in with the background, especially if you remove the borders. In addition, I've added the ability to add custom highlights in the config, in case similar issues to this exist.


Pocco81:
Thank you :+1:

PR #161, opened 2022-06-25
fix: native_lsp highlights in sign column #160

### LINKED ISSUE #160: Gitsign icons have weird background
Hi!
First off - really great theme, thanks for great efforts!
I have a question / issue that I am not sure how to deal with. Gitsign icons have slightly different background in `nvim-tree` and in the buffers as well, making it look like broken a bit.
Would you have any advice on that?
image
image

Thanks!

ps-143:
Fixes #160 . This is my first PR :)


vinibispo:
@Pocco81 I think it's a great PR

PR #162, opened 2022-06-26
use autocmd event instead of after_loading hook

laxtiz:
When we have a lot of 3rd party plugins, everyone will do a hook and wrap what's already there, it's too complicated.

why not use events?


Pocco81:
Good idea! thanks :)

PR #168, opened 2022-07-10
docs(README): fix coc tab spacing

anachronic:
There's a small misalignment that can throw some people off (myself included). This PR uses tab instead of spaces to indent coc.nvim config

image


Pocco81:
thank you!

PR #171, opened 2022-07-12
fix(syntax): Fix missing `,` in config.lua

NICHTJ3:
Missing comma in config causing catppuccin to fail to load


Pocco81:
Thanks but it's already been fixed :)

Issue #173, opened 2022-07-12
Broken on latest neovim nightly


hahuang65:
With latest neovim-nightly (NVIM v0.8.0-dev+595-g95c65a6b2)

starting up will toss an error:

```
packer.nvim: Error running config for catppuccin: /home/hao/.config/nvim/lua/plu
gin/catppuccin.lua:62: Vim(lua):E5108: Error executing lua ...ck/packer/start/ca
tppuccin/lua/catppuccin/utils/util.lua:25: 'blend' must be an integer
stack traceback:
^I[C]: in function 'nvim_set_hl'
^I...ck/packer/start/catppuccin/lua/catppuccin/utils/util.lua:25: in function 'h
ighlight'
^I...ck/packer/start/catppuccin/lua/catppuccin/utils/util.lua:44: in function 's
yntax'
^I...ck/packer/start/catppuccin/lua/catppuccin/utils/util.lua:89: in function 'l
oad'
^I...ite/pack/packer/start/catppuccin/lua/catppuccin/main.lua:19: in function 'l
oad'
^I...ite/pack/packer/start/catppuccin/lua/catppuccin/main.lua:32: in function 'm
ain'
^I...ite/pack/packer/start/catppuccin/lua/catppuccin/init.lua:4: in function 'lo
ad'
^I[string ":lua"]:1: in main chunk
^I[C]: in function 'cmd'
^I/home/hao/.config/nvim/lua/plugin/catppuccin.lua:62: in main chunk
^I[C]: in function 'require'
^I[string "..."]: in function <[string "..."]:0>
^I[C]: in function 'pcall'
```

nullchilly:
Note that this isn't a new behavior on neovim nightly.

This was merged before 0.7 was released https://github.com/neovim/neovim/pull/17516

PR #174, opened 2022-07-12
fix(highlight): Fix blend option

nullchilly:
I downloaded neovim 0.5 appimage and blend has always been integer value. No idea why is `nocombine` from `cterm` is here

Closes https://github.com/catppuccin/nvim/issues/173


Pocco81:
Thank you :)

PR #175, opened 2022-07-12
fix(g.colors_name): Set colorscheme name

nullchilly:
Fix:

https://github.com/catppuccin/nvim/commit/482de1a974f4405a70a7802341c9ada463d6a856

- `:colorscheme` no longer return catppuccin

- lualine `theme = "auto"` doesn't work


datamonsterr:
Exactly what I want to add! I wake up, PackerSync then all my custom highlights can't be applied because I custom it base on `vim.g.colors_name`

Issue #178, opened 2022-07-12
MIGRATION GUIDE (follow if your config magically broke!)


Pocco81:
Heya! After #169 was opened a bunch of changes have taken place. Some of them have arrived earlier to the `main` branch, however most still live under the `dev` branch. Not all of the interfaces were changed but the ones that were might break your config if not updated.

So, here is this issue to aid in those situations. In the meantime, while I finish reworking the whole plugin you may stay in the `main` branch and follow this issue to be up-to-date with everything.

> Note 1: the `dev` branch can be merged into `main` at any time! beware!
> Note 2: some new features are subject to change while they are still under development

## Heads-ups!

1. #177: a commit was amended and because of that I had force push it to the repo. If your plugin fails to update (because most plugin managers mishandle this kind of thing): delete the plugin's directory and reinstall it.
2. #176: the way highlights are handled changed, most notably the `style` property of the highlights. Instead of using commas to separate them (e.g. `style = "italic,bold,underline"`), please use a Lua table. Note that if you don't want any highlights just pass an empty table `{}`. For the average user's conf this would mean that:
```lua
	styles = {
		comments = { "italic" },
		conditionals = { "italic" },
		loops = {},
	},
```

This also applies to custom highlights. (Note: in this case you may either pass an empty table to `style` for none, or not pass `style` in the first place)

```lua
custom_highlights = {
        TSConstBuiltin = { fg = cp.peach, style = {} },
        TSConstant = { fg = cp.sky },
        TSComment = { fg = cp.surface2, style = { "italic" } }
    }
```
3. c41a627: the code-base was refactored and now:
+ `palettes`: holds the palettes and the interface to fetch them
+ `utils`: general utilities for handling data and color properties
+ `groups`: holds the highlights groups and integrations
+ `lib`: main code. Holds the `compiler`, `highlighter`, `mapper`, `ui` and `hsluv`

So please don't forget to change your calls to any of Catppuccin's modules!

4. This ain't much of a breaking feature but since the new update users are now able to pre-compile their configs to improve startup speed. [Check it out](https://github.com/catppuccin/nvim/#compile)!

- Startuptime stat: 0.7ms
```
012.793  000.096  000.096: require('catppuccin')
012.835  000.035  000.035: require('catppuccin.config')
013.356  000.005  000.005: require('vim.F')
014.246  000.669  000.669: sourcing /home/nullchilly/.local/share/nvim/site/pack/packer/start/theme/colors/catppuccin.lua
```

[...]

PR #179, opened 2022-07-13
refactor(highlight): Deprecating style = string. Use style = table

nullchilly:
The document is already out before this pr D:

https://github.com/catppuccin/nvim/issues/178


Pocco81 (approved):
looking good.

PR #196, opened 2022-07-15
clarify and fix README description of dim_inactive in config

jorgebef:
dim_inactive seems to be incorrectly explained and example provided in README.
It could mislead many people since it does result in a broken config.


nullchilly:
Can you re-format the compile section too?

I am very bad at writing readme 😅

PR #218, opened 2022-07-30
Fix typo in readme

tom-draper:
Noticed a missing comma in two code snippet examples in the readme that will cause an error.


nullchilly:
Thanks!

PR #226, opened 2022-08-04
Adding support to nvim-navic

augustocdias:
I added support to nvim-navic... I've tried to base the colors on the cmp and on treesitter. 

One issue that I'm not sure how to solve is the background. I thought putting `NONE` on all of them would make it transparent, but it is not happening, when I set it in lualine for example I'm getting the `base` background instead. Perhaps you could point me to how to solve this :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ This port of Catppuccin is special because it was the first one and the one that
     -   [VimWiki](https://github.com/vimwiki/vimwiki)
     -   [Leap.nvim](https://github.com/ggandor/leap.nvim)
     -   [Nvim-navic](https://github.com/SmiteshP/nvim-navic)
-    -   [vim-clap](https://github.com/liuchengxu/vim-clap)
+    -   [Vim-clap](https://github.com/liuchengxu/vim-clap)
 
 # Installation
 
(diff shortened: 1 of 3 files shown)


nullchilly:
What is that output of `:hi NavicIconsFile` ?

augustocdias:
> What is the output of `:hi NavicIconsFile` ?

`NavicIconsFile xxx guifg=#1e66f5`

nullchilly:
Hmm that's weird, maybe nvim-navic is overriding catppuccin. Can you investigate the loading order?

nullchilly:
@augustocdias If you don't mind me commit directly to your repo

I will investigate for you :)

nullchilly:
@augustocdias You have to enable highlight in nvim-navic

```lua
require("nvim-navic").setup {
	highlight = true,
}
```

augustocdias:
@nullchilly  I did it. It is highlighting correctly, just the background that is not working.

nullchilly:
@augustocdias Can you send screenshot? it was working for me...

augustocdias:
Note that the background is the base instead of the mantle which is the one in this lualine section

image

nullchilly:
You need to add custom highlight in that lualine section, it has no affiliation with nvim-navic

augustocdias:
I set them in the integration... Wouldn't that be enough?

nullchilly:
@augustocdias Try `:hi NavicIconsFile guibg="NONE"`

Then `:hi NavicIconsFile`

You can see that vim doesn't register guibg

augustocdias:
So even if I set it through lualine, it is not respected. Should we add a config to allow setting a custom bg for navic instead of always setting it to NONE?

something like 
```
navic = {
    enabled = true,
    custom_bg = '#000000'
}
```

PR #235, opened 2022-08-07
fix(neogit): improve Neogit colors

mizlan:
## before

Screen Shot 2022-08-06 at 5 52 43 PM

## after

Screen Shot 2022-08-06 at 5 40 42 PM

### DIFF
--- a/lua/catppuccin/groups/integrations/neogit.lua
+++ b/lua/catppuccin/groups/integrations/neogit.lua
@@ -18,24 +18,24 @@ function M.get()
 			fg = cp.text,
 		},
 		NeogitDiffDeleteHighlight = {
-			bg = ucolors.darken(cp.red, 0.095, cp.base),
-			fg = cp.red,
+			bg = ucolors.darken(cp.red, 0.345, cp.base),
+			fg = ucolors.lighten(cp.red, 0.850, cp.text),
 		},
 		NeogitDiffAddHighlight = {
+			bg = ucolors.darken(cp.green, 0.345, cp.base),
+			fg = ucolors.lighten(cp.green, 0.850, cp.text),
+		},
+		NeogitDiffDelete = {
+			bg = ucolors.darken(cp.red, 0.095, cp.base),
+			fg = ucolors.darken(cp.red, 0.800, cp.base),
+		},
+		NeogitDiffAdd = {
 			bg = ucolors.darken(cp.green, 0.095, cp.base),
-			fg = cp.green,
+			fg = ucolors.darken(cp.green, 0.800, cp.base),
 		},
 		NeogitCommitViewHeader = {
 			bg = ucolors.darken(cp.blue, 0.300, cp.base),
 			fg = ucolors.lighten(cp.blue, 0.800, cp.text),
-		},
-		NeogitDiffDelete = {
-			bg = ucolors.darken(cp.red, 0.125, cp.base),
-			fg = ucolors.darken(cp.red, 0.445, cp.base),
-		},
-		NeogitDiffAdd = {
-			bg = ucolors.darken(cp.green, 0.125, cp.base),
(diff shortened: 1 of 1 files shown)


mizlan:
As an active Neogit user, I find the existing colors intolerable so I created this new set of highlights that takes advantage of `ucolors.darken` a lot!

nullchilly:
Thank! I am still using vim-fugitive
I guess its time to try neogit out :)

nullchilly:
@mizlan Don't you think it's still a bit hard to see in NeogitCommitView?
image

lua/catppuccin/groups/integrations/neogit.lua
```diff
@@ -1,14 +1,30 @@
...
 		NeogitRemote = { fg = cp.pink },
-		NeogitHunkHeader = { bg = cp.blue, fg = cp.text },
-		NeogitHunkHeaderHighlight = { bg = cp.base, fg = cp.blue },
-		NeogitDiffContextHighlight = { bg = cp.mantle, fg = cp.overlay2 },
-		NeogitDiffDeleteHighlight = { bg = cp.mantle, fg = cp.red },
-		NeogitDiffAddHighlight = { bg = cp.mantle, fg = cp.green },
+		NeogitHunkHeader = {
+			bg = ucolors.darken(cp.blue, 0.095, cp.base),
+			fg = ucolors.darken(cp.blue, 0.5, cp.base),
+		},
```
nullchilly, line 11: Can you add NeogitCommitViewHeader?

mizlan: got it.

mizlan:
Also see this: https://github.com/TimUntersberger/neogit/issues/330

Will work on this, will also probably PR Neogit to get customization on the other hunks (I want to dim unfocused hunk's context lines)

mizlan:
Screen Shot 2022-08-06 at 6 38 00 PM

Foreground also turns weird blue color in Commit view mode, will try to fix as well

mizlan:
This is the best I can do for now, since I dont know why it's blue. Fixed formatting, and added some more hlgroups. Take a look:
Screen Shot 2022-08-06 at 8 48 05 PM
Screen Shot 2022-08-06 at 8 48 29 PM

nullchilly:
@mizlan It's too dim for me though?
image

mizlan:
Yeah, thats a bit of a problem. However, the goal is to fix that in Neogit itself so the hunk under cursor uses the other set of highlights. These, then, would be the "inactive" hunks. I think for now this is an alright approach.

nullchilly:
@mizlan So we are missing the ActiveHunk

Why not just decrease the dim in the meantime and open another pr later when InactiveHunk is fixed?

mizlan:
Good question. Maybe my workflow is different than others but I spend 99% of my time in the status buffer, where the dimmed text matters and is significantly helpful. I rarely view existing commits or browse diffs in other forms in Neogit. I cannot say the same for others' workflows but I'll be happy to change it if you prefer. I could leave those commented out with a note to re-enable them after Neogit is updated.

nullchilly:
Yeah that sounds good

Remember we try to not be opinionated as much as possible

Personally reference is done via custom_highlight :)

mizlan:
Oh. I've got an idea. I'll use a more vibrant bg for "active" hunks right now. Will be easily visibly distinguishable while allowing both active and inactive to be readable.

mizlan:
Screen Shot 2022-08-06 at 9 26 39 PM
Quite fond of this setup: both are legible but one is grabs my attention more.

nullchilly:
@mizlan Seems good!

We can merge this now if you want?

mizlan:
Sure, go ahead! Can't find anything else for now.

PR #246, opened 2022-08-19
fix & feat(bufferline): Correct colouring on errors, warnings, infos and hints and option to disable italics and or bolds.

RAV64:
This pr adds options to config to remove italics/bolds from bufferline with options so there's not a need to create separate custom rule for every single one. It also adds correct colouring to errors, warnings, infos and hints.


RAV64:
[d98dbdd](https://github.com/catppuccin/nvim/pull/246/commits/d98dbdde62d3dadb49a4674583cde5a36b344432) also defaulted bolding icons off since it just looked blurry before. Can be enabled with custom rules but I doubt people want it.

Pocco81:
thanks for the PR!

nullchilly:
@Pocco81 Please add backward compatibility to `bufferline = true`

Pocco81:
@nullchilly that should already be working: https://github.com/catppuccin/nvim/blob/dev/lua/catppuccin/lib/mapper.lua#L11-L19

PR #249, opened 2022-08-24
Improving background of navic to respect transparency options or use default

augustocdias:


nullchilly:
Shouldn't navic already respect transparency with `bg = "NONE"` though? Am I missing something

augustocdias:
As I explained before it doesn't work when you put it inside lualine for example. And setting the colors in lualine doesn't work either.

nullchilly:
Please add a `custom_bg` option in that case, we don't want to assume mantle is what people want for their lualine.

Also remember the [backward compatibility](https://github.com/catppuccin/nvim/commit/3a5f324694b0bef9206bc85c10bbf14817f81f98)

augustocdias:
Sorry I was going to do it this weekend.

PR #265, opened 2022-08-28
feat: Add nvim-treesitter-context integration

### LINKED ISSUE #264: feat: Add nvim-treesitter-context integration

mslalith:
Closes #264


nullchilly:
Please use the highlight groups mentioned here https://github.com/nvim-treesitter/nvim-treesitter-context#appearance instead of the fallback group

mslalith:
@nullchilly updated. Please have a look

nullchilly:
If too many treesitter plugins pop up I guess we should merge them into one table, like this for example:

```lua
treesitter = {
  enabled = true,
  context = true,
  rainbow = true
}
```

PR #267, opened 2022-08-29
Fix: TS-Rainbow

RAV64:
Removing unnecessary background highlighting since parenthesis should always have same background as the current background by default.

Before:

image

After:

image

### DIFF
--- a/lua/catppuccin/groups/integrations/ts_rainbow.lua
+++ b/lua/catppuccin/groups/integrations/ts_rainbow.lua
@@ -1,20 +1,14 @@
 local M = {}
 
 function M.get()
-	local cnf = require("catppuccin.config").options
-	local transparent_background = cnf.transparent_background
-	local bg_highlight = (transparent_background and cnf.dim_inactive.enabled and cp.dim)
-		or (transparent_background and "NONE")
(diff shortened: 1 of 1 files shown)


gotgenes:
@RAV64 Thanks for your contribution! It looks like your PR is failing on the StyLua check.

Do you have access to [StyLua](https://github.com/JohnnyMorganz/StyLua) on your system? If so, could you please run

```
stylua lua/catppuccin/groups/integrations/ts_rainbow.lua
```

and commit the changes?

The indentation in this project is with tabs not spaces, which is what I believe is why the StyLua check is failing.

RAV64:
I was fairly certain that check was just bugged so I left it at that. Shouldn't there be stylua command ran on the project before every push? Hope It's good now :)

RAV64:
Also there seems to be typo in the .editorconfig @gotgenes. Can I add commit that changes `indent_style = tabs` -> `indent_style = tab` since "tabs" isn't possible value for indent_style as stated here: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_style. With this change my first commits should have passed.

gotgenes:
Great catch on the editorconfig setting. Let's correct that as a separate PR.

Thanks for this contribution!

PR #268, opened 2022-08-30
Fix: editorconfig

RAV64:
Changes `indent_style = tabs` -> `indent_style = tab` since "tabs" isn't possible value for indent_style as stated here: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_style.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ root = true
 end_of_line = lf
 charset = utf-8
 insert_final_newline = true
-indent_style = tabs
+indent_style = tab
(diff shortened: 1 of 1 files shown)


nekowinston:
Oh yeah, that's a hangover from the https://github.com/catppuccin/template repo, which got fixed with https://github.com/catppuccin/template/commit/11fa5d35a8c83c2bfb565ae5c28d31d70a53b527

LGTM, think this should impact nothing else regarding the port itself.

PR #275, opened 2022-09-03
chore(readme): Sort integrations in case-insensitive alphabetical order.

gotgenes:
As a user, I think it's easier to look through the integrations in alphabetical order rather than chronological order in which support was added in catppuccin, especially if you're new to the colorscheme, or wondering, "Does this support my favorite plugins?"


gotgenes:
Hmm, something went wrong with this edit; there's a lot more content missing than should be. Aborting and trying again later.

PR #279, opened 2022-09-07
Add pounce.nvim integration

pulcinello:
[Pounce](https://github.com/rlane/pounce.nvim) is a motion plugin which can do incremental fuzzy search. Here are pictures show the final result.
before:
Screen Shot 2022-09-08 at 04 02 05
after:
Screen Shot 2022-09-08 at 04 03 15


gotgenes (approved):

gotgenes:
Thanks for your contribution! For what it's worth, for documentation, you only need to update the `README.md` right now. The `catppuccin.txt` file is automatically generated when we merge, but this is not a problem.

PR #284, opened 2022-09-22
fix(CI): Broken formatting workflow (ELF not found).

Jint-lzxy:
This commit fixed the problem with formatting CI. Currently incorrect artifact was chosen due to the introduction of `linux-aarch64` artifacts of latest `stylua-action` _(a.k.a, `v1.1.0`)_, which would cause "ELF not found" related errors.
> Ref: https://github.com/JohnnyMorganz/stylua-action/releases/tag/v1.1.1


nullchilly:
Thanks you 🥲

PR #295, opened 2022-10-08
Add FAQ Section for TrueColor

jdevries3133:
macOS's "Terminal.app" does not support TrueColor. It can be tricky to identify
what's going wrong when users encounter this issue. Hopefully this will point
them in the right direction by identifying that Terminal.app doesn't have the
color support that catppuccin requires.

### DIFF
--- a/README.md
+++ b/README.md
@@ -830,11 +830,17 @@ require("nvim-treesitter.configs").setup {
 }
 ```
 
-## Setup for macOS "Terminal.app"
+## Colors doesn't match preview screenshots
 
-Trick question - not possible! Catppuccin requires TrueColor, which macOS's
-native "Terminal.app" does not support. Consider using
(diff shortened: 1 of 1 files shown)


nullchilly:
Thanks you! I think this is the case for all terminal that does not support true color or isn't enabled by default

jdevries3133:
Hm, I do understand that. I want to keep the keywords because I think they will get Ctrl + f hits. I know the presence of these certainly would have saved me some trouble. To that end, though, I'm ok with using that title but keeping the keywords in the body. 

What do you think? Do you think there might be additional keywords that might be worthwhile to include in the same vein?

nullchilly:
Ah I didn't see your comment 1 minute ago

nullchilly:
image

jdevries3133:
LGTM, thanks!

PR #309, opened 2022-10-19
docs: added images to repo + converted to webp

isabelincorp:
@nullchilly Opinions? Note all repos except for this one and two others has an assets folder.


nullchilly:
assets/ is a dead no for vim plugins, It existed in the past and I removed it.

Images make the git repo at least 3x heavier when cloning for seemingly no reason.

sgoudham:
Just a comment from me. 

In all fairness, the changes proposed here are fairly valid for catppuccin. Sure, the images are pulled down for "seemingly no reason" but it's not the 1990's anymore where kilobytes matter in the grand scheme of things. 

Unfortunately, rejecting this type of (valid docs) change might prevent future contributors wanting to contribute. Hope you might reconsider your stance on this at some point @nullchilly.

nullchilly:
It's 300KB, not a few. If every plugin does this it becomes 3MB, someone will blame packer for timing out.

> Unfortunately, rejecting this type of (valid docs) change might prevent future contributors wanting to contribute

I am not sure what's the problem here, you can upload your image into the pull request description itself then link markdown to it.

gotgenes:
I appreciate the intent of this change. I agree with @nullchilly, though. [The convention for Neovim plugins is to make sure assets are not stored within the repo itself.](https://www.reddit.com/r/neovim/comments/wsg5yx/attention_to_all_plugin_authors_newcomers_careful/)

nekowinston:
I'm in favour of keeping the screenshots in **a** repository, it doesn't have to be this one.

The main reason for that is that we have more control of the assets when they're stored in a repository.
- we can mirror the repository without relying on GitHub CDN
- people have local versions without GitHub CDN connections
- We can more easily use the assets on our website (still WIP atm), if they have a standardized naming scheme.

Maybe this could be an improvement for the whole org? We could have a **separate repository** in which we keep screenshots, which would solve all the points above.
However from a port maintainer/contributor perspective, it might be a bit weird to keep the screenshots updated in a dedicated repository? I'd be happy to hear more ideas how we could solve this.

isabelincorp:
> 

1. Yea I agree, order can be up to the port maintainer 
2. Made sure to remove that after looking over the whole README 
In conclusion: I won't be reopening unless we do something like what winston mentions. I think making a repo and using it for programs like these makes sense. 

Side-Note: I have barely dipped my toes into vim, so I'm still learning about how neovim plugins work.

gotgenes:
@isabelincorp I think for **most** projects, this would be a helpful, welcome change. Vim/Neovim have no native plugin packaging system, and essentially rely on `git clone` as the mechanism for obtaining plugins. That means the courteous thing for our users is to make sure to be very conscientious of what we commit to the repo, and that means omitting our assets.

In other words, thank you! I know you were trying to make an improvement.

isabelincorp:
I would want to find some way to automate a master repo, which, IMO, could work extremely well if done right. How I would want the automation to work is to take any newly transferred repo and transfer all the contents of assets into their own folder within the master repo + deal with relinking the README's.

PR #310, opened 2022-10-19
Treesitter Highlight Maps + Aerial fix

### LINKED ISSUE #307: Aerial.nvim integration
### Description

Aerial.nvim highlight groups still link to the old `TS*` highlight group syntax, which breaks with the new nvim-treesitter.

### Neovim version

```markdown
NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@Monterey
```

### Operating system and version

macOS 12.15.1

main

### Steps to reproduce

n/a

### Expected behavior

Aerial highlights work

### Actual behavior

Highlight groups cleared

### Minimal config

```Lua
-- `minimal.lua` used for reproducible configuration
-- Open with `nvim --clean -u minimal.lua`

local is_windows = vim.fn.has("win32") == 1
local function join(...)
  local sep = is_windows and "\\" or "/"
  return table.concat({ ... }, sep)
end

local root_tmp = is_windows and os.getenv("TEMP") or "/tmp"
local site_path = join(root_tmp, "nvim", "site")
local pack_path = join(site_path, "pack")
local install_path = join(pack_path, "packer", "start", "packer.nvim")
local compile_path = join(install_path, "plugin", "packer_compiled.lua")
vim.opt.packpath = site_path

local function load_plugins()
  local packer = require("packer")
  local use = packer.use
  packer.reset()
  packer.init({ compile_path = compile_path, package_root = pack_path })

  use("wbthomason/packer.nvim")
  use({ "catppuccin/nvim", as = "catppuccin" })
  -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE

  packer.install()
end

_G.load_config = function()
  -- ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
  vim.g.catppuccin_flavour = "macchiato"
  require("catppuccin").setup({
    integrations = {
      aerial = true,
    },
  })

  vim.cmd("colorscheme catppuccin")
end

if vim.fn.isdirectory(install_path) == 0 then
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
vim.cmd([[autocmd User PackerComplete ++once echo "Ready!" | lua load_config()]])
```

axieax:
This PR refactors to use Treesitter highlight group mappings, and resolves #307.


nullchilly:
Why didn you delete `TS*` highlights group despite adding a fallback to nvim 0.7?

lua/catppuccin/groups/integrations/treesitter.lua
```diff
@@ -11,213 +11,118 @@ function M.get()
...
-			typescriptTSConstructor = { fg = cp.lavender },
-
-			-- TSX (Typescript React)
-			tsxTSConstructor = { fg = cp.lavender },
-			tsxTSTagAttribute = { fg = cp.mauve },
-
-			-- cpp
-			cppTSProperty = { fg = cp.rosewater },
-
-			-- yaml
```
nullchilly, line 217: I meant restore this table so the fallback map actually works

Please test these changes on nvim 0.7.2 appimage

axieax:
My intent is to migrate to the new Treesitter highlight syntax, but also add a fallback for the old way `TS*`. This mapping refactor reduces repeated code and allows for future development to use the new syntax.

nullchilly:
@axieax You are missing the point.

You deleted the old TS* group in nvim 0.7, there is nothing to fallback into

axieax:
> @axieax You are missing the point.
> 
> You deleted the old TS* group in nvim 0.7, there is nothing to fallback into

Doesn't this simply map the new syntax to the old syntax?

https://github.com/axieax/catppuccin-nvim/blob/9c5f806a04b8d40dafdc2c877703071ce7323381/lua/catppuccin/groups/integrations/treesitter.lua#L115-L123

nullchilly:
@axieax Ah yeah, you're right 😅

Reviewing on phone is hard

@gotgenes Would you mind testing this pull request?

nullchilly:
@axieax [Wait...](https://github.com/nvim-treesitter/nvim-treesitter/commit/42ab95d5e11f247c6f0c8f5181b02e816caa4a4f#), its actually still broken as treesitter no longer support `TS*` highlights. Why do we care so much about legacy fallback to `TS*` when nvim-treesitter itself now requires nvim 0.8?

axieax:
Yeah I get what you mean. Do you think it's better if we remove the mapping altogether and only support the updated syntax?

nullchilly:
@axieax Maybe something like this?

```lua
if vim.treesitter.highlighter.hl_map then
  vim.notify([[Catppuccin (info):
nvim-treesitter integration requires neovim 0.8
If you want to stay on nvim 0.7 disable the integration or pin catppuccin tag to v0.2.4 and nvim-treesitter to ...
]], "info")
```

axieax:
Should be ready now :))

nullchilly (changes requested):

axieax:
Addressed!

PR #330, opened 2022-10-26
Added Felines new provider: macro and search count

ilias777:
Feline added new providers for macro recordings and search count.
This is useful if `cmdline=0`. By pressing `qa...` it's shows the recording next to the *position* component. Same for the search count. 
If you don't like the position of the two components, feel free to position them elsewhere.

macro

search_count

### DIFF
--- a/lua/catppuccin/groups/integrations/feline.lua
+++ b/lua/catppuccin/groups/integrations/feline.lua
@@ -269,6 +269,7 @@ function M.get()
 	-- macro
 	components.active[1][12] = {
 		provider = "macro",
+		enabled = function() return vim.api.nvim_get_option "cmdheight" == 0 end,
 		hl = {
 			fg = sett.extras,
 			bg = sett.bkg,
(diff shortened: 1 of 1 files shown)


nullchilly:
It should be enabled by default only if cmdheight=0

arpangreat:
@ilias777 what font is that ?

ilias777:
> It should be enabled by default only if cmdheight=0

Yes you are right!
Now is finish! The `macro` and the `search_count` are enabled only if `cmdheight=0`

ilias777:
> @ilias777 what font is that ?

The font is [Cascadia Code](https://github.com/microsoft/cascadia-code)
This font is amazing!!!
If you are on Kitty Terminal, make sure you are enabled the font features of Cascadia Code in the `kitty.conf` file like so:
```
## CASCADIA CODE ##
font_family Cascadia Code
bold_font Cascadia Code Bold
italic_font Cascadia Code Italic
bold_italic_font Cascadia Code Bold Italic

## FONT FEATURES FOR CASCADIA CODE ##
font_features CascadiaCode-Italic calt +ss01 +ss02 +ss03 +ss04 +ss05 +ss07 +ss08 +zero +onum
```
This make the italic version even better 😉

nullchilly:
cmdheight ftw ❤️

PR #332, opened 2022-10-31
fix(bufferline): Respect gui styles

hanxi:
The example in readme is set no italic .


README.md
```diff
@@ -503,7 +503,7 @@ Configurations are self-explanatory, see `:h bufferline-highlights` for detailed
 local mocha = require("catppuccin.palettes").get_palette "mocha"
 bufferline.setup {
     highlights = require("catppuccin.groups.integrations.bufferline").get {
-        styles = { "italic", "bold" },
```
nullchilly, line 506: You can use `style = { "bold" }` to disable italic. I don't think this is a bug?

The reason why we have this style array is historical reason

PR #337, opened 2022-11-12
fix(lsp): LspInfo border color

WilliamHsieh:
Fix the LspInfo border color.

Before:
Screenshot 2022-11-12 at 15 31 44

After:
Screenshot 2022-11-12 at 15 32 03


Pocco81:
Good change. Fits more with the overall aesthetics of the color scheme and its more prevalent colors.

PR #340, opened 2022-11-17
feat: add vim support

nullchilly:


EdenEast:
Awesome to see catppuccin support vim as well! Great work @nullchilly!

nullchilly:
@EdenEast Thanks you :)

PR #346, opened 2022-11-19
fix(leap): respect transparency

emxxjnm:
image


nullchilly:
Please keep the `style = {}`

PR #352, opened 2022-11-27
fix(nix): no longer stuck in 1969

nullchilly:
cc @EdenEast

1. Hash user and git file name when `mtime = 1` (Which means nix is presented)

2. viml's getftime is faster than uv.fs_stat

```lua
local now = vim.loop.hrtime
local start = now()
for _ = 1, 10000 do
	vim.fn.getftime "/home/nullchilly/.cache/nvim/catppuccin/date.txt"
end
print("getftime  " .. (now() - start) / 1000000 .. "ms")

start = now()
for _ = 1, 10000 do
	vim.loop.fs_stat "/home/nullchilly/.cache/nvim/catppuccin/date.txt"
end
print("luajit    " .. (now() - start) / 1000000 .. "ms")
```

```lua
getftime  7.314094ms                                                                           
luajit    16.374475ms
```

Closes #348

PR #353, opened 2022-11-27
fix(compile): .git doesn't exist on nixos

nullchilly:
cc @stasjok #352

I will also look into the re-compile issue, sorry..

### DIFF
--- a/lua/catppuccin/init.lua
+++ b/lua/catppuccin/init.lua
@@ -120,9 +120,9 @@ function M.setup(user_conf)
 	end
 
 	local config_path = debug.getinfo(2).source:sub(2) -- Get user config path
-	local git_path = debug.getinfo(1).source:sub(2, -24) .. ".git" .. M.path_sep .. "ORIG_HEAD" -- Get git path
+	local git_path = debug.getinfo(1).source:sub(2, -24) -- Get git path
 	local config = vim.fn.getftime(config_path) -- getftime is 2 time faster than fs_stat, for benchmark see #352
-	local git = vim.fn.getftime(git_path) -- Parsed config & git stat
+	local git = vim.fn.getftime(git_path .. ".git" .. M.path_sep .. "ORIG_HEAD") -- Parsed config & git stat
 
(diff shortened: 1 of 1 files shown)


lua/catppuccin/init.lua
```diff
@@ -120,9 +120,9 @@ function M.setup(user_conf)
 	end
 
 	local config_path = debug.getinfo(2).source:sub(2) -- Get user config path
-	local git_path = debug.getinfo(1).source:sub(2, -24) .. ".git" .. M.path_sep .. "ORIG_HEAD" -- Get git path
+	local git_path = debug.getinfo(1).source:sub(2, -24) -- Get git path
 	local config = vim.fn.getftime(config_path) -- getftime is 2 time faster than fs_stat, for benchmark see #352
-	local git = vim.fn.getftime(git_path) -- Parsed config & git stat
+	local git = vim.fn.getftime(git_path .. ".git" .. M.path_sep .. "ORIG_HEAD") -- Parsed config & git stat
 
 	local cur_date = (config == 1 and config_path or config) .. (git == 1 and git_path or git) -- nix mtime is always 1 so cache path instead
```
stasjok, line 127:
```suggestion
	local cur_date = (config == 1 and config_path or config) .. (git == -1 and git_path or git) -- nix mtime is always 1 so cache path instead
```

nullchilly:
@stasjok I think I should just disable the fragile user config mtime to be honest

nullchilly:
Because it would need three seperate files cache at this point

`date`, `git` and `config`

stasjok:
Latest version has updated `date.txt` with:
```
1669531780/nix/store/xi9ym8i2zlxhm41pdqdcmxx1ynh5b359-vim-pack-dir/pack/myNeovimPackages/start/catppuccin-nvim/.git/ORIG_HEAD
```
So it should work. `*_compiled.lua` files aren't updated, but it's a different problem not related to nix.

nullchilly:
Alright, I will merge this first, then open another PR to fix that different problem

PR #355, opened 2022-11-28
Fix: @type.qualifier color pink according to examples

robmeijerink:
Ever since [this commit](https://github.com/nvim-treesitter/nvim-treesitter/commit/0f866c15b4e77406c3eb4110e7ddbabce52ccaa1) in treesitter's git repo. The colors in my PHP files look off. Private / protected / public keywords in classes are now yellow and do not really fit its surrounding colors. I am not sure if this yellow is the intended color for Catppuccin's theme, because if I look at examples / images these keywords are usually pink; for example in [this image](https://preview.redd.it/2zwg74ym3n381.png?width=1327&format=png&auto=webp&s=f2a4e2f466be53bed6deee97de0831b452863a0a). Pink fits better, because it will not make these keywords stand out as much.

I have the impression that this pink color is the intended one for these keywords. If I am mistaken and the keywords are yellow on purpose, then close the PR, and I will change it myself in the config.


nullchilly:
Please add it as php only, like this: `["@type.qualifier.php"]`

robmeijerink:
@nullchilly Thank your for your quick reply. I added it as PHP only.

PR #358, opened 2022-11-29
fix(syntax): lack of foreground text for `@text.todo`

jackfranklin:
`@text.todo` was lacking a foreground color so it would take the colour from the comment - this in the latte theme lead to a pretty nasty grey-on-orange theme. Compared to the other types like `@text.danger`, it seems we're missing the `fg` property which this PR sets.

As a drive-by change, I also added a missing brace in the example code for overrides in `README.md`.

Here's an example of the issue:

Screenshot 2022-11-29 at 16 06 24

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ require("catppuccin").setup {
         return {
             Comment = { fg = colors.flamingo },
             ["@constant.builtin"] = { fg = colors.peach, style = {} },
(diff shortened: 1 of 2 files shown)


nullchilly:
Ah I have to `:TSInstall comment` to enable this feature, no wonder why TODO wasn't highlighting

Thank you for this!

jackfranklin:
You're welcome - thanks for the theme! :)

PR #360, opened 2022-12-03
fix(neotest): consistency with original theme

igorlfs:
Original neotest's theme:
image
Current catppuccin integration:
image
With this PR:
image


igorlfs:
Wow, thanks for the fast merge!

PR #370, opened 2022-12-16
feat(lualine): add a specific color for the terminal mode

ofseed:
The default color of the `terminal` mode in lualine is the same as the `normal` mode,
specify a color for the terminal mode to recognize whether the mode is either `terminal`
or `normal`.


ofseed:
`terminal` mode is another `insert` mode in my view, so I make the highlight of them the same.

PR #376, opened 2022-12-31
feat(barbecue): add initial support

CnTeng:
[barbecue.nvim](https://github.com/utilyre/barbecue.nvim) is a VS Code like winbar for Neovim. This is how it looks like
20221231_18h00m12s_grim


CnTeng:
https://github.com/catppuccin/nvim/pull/377 maybe is better

PR #377, opened 2022-12-31
feat(barbecue): add barbecue

ofseed:
Coincidentally I found #376  after finishing it,
However, there are some differences between mine and his. 
The main difference is that I use Conceal for the file path and it uses Normal,
which is a part that is unique to barbecue and not available in navic.
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,7 @@ Below is a list of supported plugins and their corresponding integration module.
 | ------------------------------------------------------------------------------------- | ------------------- |
 | [aerial.nvim](https://github.com/stevearc/aerial.nvim)                                | aerial              |
 | [barbar.nvim](https://github.com/romgrk/barbar.nvim)                                  | barbar              |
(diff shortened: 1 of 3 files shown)


nullchilly:
Personally I think the folder path is a bit dim
@CnTeng thoughts?

CnTeng:
I just copy the color from navic, maybe his choose is better.

nullchilly:
This version's dirname is too dim, we should cherry picking from @CnTeng 's PR

ofseed:
In fact, the dimed color is in recommended in barbecue's documentation. I think I can add an option to this pr for the user to decide this color later.

ofseed:
@nullchilly Done. Now you can customize the dirname color yourself.

ofseed:
And the new version of barbecue, which could generate colorscheme automatically, dims the dirname. So making it a default option is also a good choice.

PR #382, opened 2023-01-07
LICENSE: Change year to 2023

ivabus:
Happy New Year!


gotgenes:
It is a common misconception that the copyright year must be updated every year. [This is a helpful reference.](https://opensource.stackexchange.com/a/6393)

ivabus:
thanks

PR #388, opened 2023-01-20
fix(load): flavour mechanism

WilliamHsieh:
Tested:
- Set colorscheme when g:colors_name is present (Hence the lock check)
  - whether g:colors_name == "catppuccin" initially
- Respect o:background
- Change flavour
- Respect User default flavour

Resolve: https://github.com/catppuccin/nvim/pull/380

### DIFF
--- a/lua/catppuccin/init.lua
+++ b/lua/catppuccin/init.lua
@@ -80,17 +80,14 @@ function M.compile()
 	M.flavour = _flavour -- Restore user flavour after compile
 end
 
----@param default? string
 local function get_flavour(default)
 	local flavour
 	if default then
 		flavour = default
-	elseif vim.g.colors_name == "catppuccin" then
-		-- after first time load
+	elseif vim.g.colors_name == "catppuccin" then -- after first time load
 		flavour = M.options.background[is_vim and vim.eval "&background" or vim.o.background]
 	else
-		-- first time load
-		flavour = M.flavour
+		flavour = M.flavour -- first time load
 	end
 
 	if flavour and not M.flavours[flavour] then
@@ -108,7 +105,6 @@ end
 
 local lock = false -- Avoid g:colors_name reloading
 
----@param flavour? string
 function M.load(flavour)
 	if lock then return end
 	M.flavour = get_flavour(flavour)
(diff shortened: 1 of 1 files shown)


lua/catppuccin/init.lua
```diff
@@ -80,20 +80,44 @@ function M.compile()
...
+				flavour
+			),
+			vim.log.levels.ERROR
+		)
+		flavour = "mocha"
+	end
+
+	if flavour then
+		return flavour
+	elseif vim.g.colors_name and vim.g.colors_name == "catppuccin" then
```
nullchilly: Shouldn't 

```lua
elseif vim.g.colors_name == "catppuccin" then
```

be sufficient

lua/catppuccin/init.lua
```diff
@@ -80,20 +80,46 @@ function M.compile()
...
+
+	if flavour and not M.flavours[flavour] then
+		vim.notify(
+			string.format(
+				"Catppuccin (error): Invalid flavour '%s', flavour must be 'latte', 'frappe', 'macchiato' or 'mocha'",
+				flavour
+			),
+			vim.log.levels.ERROR
+		)
+		flavour = nil
```
nullchilly, line 104:
```lua
flavour = "mocha"
```

I think assign mocha right here make more sense

WilliamHsieh:
Sorry for the noise, it's ready for review now.

nullchilly (changes requested):

lua/catppuccin/init.lua
WilliamHsieh: you're right, didn't know lua is able to compare nil directly

lua/catppuccin/init.lua
WilliamHsieh: `M.flavour` might be nil on the initial setup, so the following line
```
return flavour or "mocha"
```
is necessary.
Of course, I can change this line to `flavour = "mocha"`, but I would like to prevent the duplication of writing "mocha" twice:)
What do you think?

nullchilly: Ah sure 😅

WilliamHsieh:
fixed

nullchilly:
To be honest I wanted to deprecate `:Catppuccin` long ago

However we lose stars everytime I make breaking changes so we keep backwards compatibility forever now 😄

nullchilly (approved):

PR #392, opened 2023-01-29
feat(lspsaga): Add support for latest highlight groups

Jint-lzxy:
This commit makes `catppuccin` compatible with `lspsaga`@v0.2.4. It added support for the latest highlight groups.

Major highlight groups have been adapted to stay harmonic with other integrations _(to the best of my ability)_. The default values for some highlight groups are just ok, so I didn't append them to the list.

<sup>Minor fixes:</sup>
<sup>\* Custom colors are no longer supported, so they have been removed.</sup>

### DIFF
--- a/doc/catppuccin.txt
+++ b/doc/catppuccin.txt
@@ -707,7 +707,6 @@ For custom Lsp Kind Icon and Color
 >lua
     require("lspsaga").setup {
         ui = {
-            colors = require("catppuccin.groups.integrations.lsp_saga").custom_colors(),
             kind = require("catppuccin.groups.integrations.lsp_saga").custom_kind(),
         },
     }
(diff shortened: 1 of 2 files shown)


VincentBerthier:
Hi,
I have a problem with this update regarding lsp_saga integration
custom_colors (the function we're supposed to call from groups/integrations/lsg_saga.lua) is a nil value.

Replacing it with the get() method (which seems to be its new name) gives an error too: 
catppuccin/groups/integrations/lsp_saga.lua:41: attempt to index global 'C' (a nil value)

Did I fail in configuring something (which is entirely possible since I'm switching to lazy), or is something really wrong here?

nullchilly:
`colors` option was removed upstream so `custom_colors` is no longer needed

@Jint-lzxy Maybe we can keep backwards compatibility?

VincentBerthier:
Oh, if colors is not needed then don't worry about it, I'll just remove that line, no worries!

PR #403, opened 2023-02-12
fix(transparent): winblend and popup

### LINKED ISSUE #401: Revert transparent background when `winblend`/`pumblend` option is set?
When the `transparent_background` is set, setting `winblend` and/or `pumblend` would cause the bg to set to black. As a result, I have to override all the highlight that is set to `None` (including the integration's colors).

Perhaps just add a condition to check if `winblend` and/or `pumblend` value is set for the respective highlight color.

**Describe alternatives you've considered**
Manually overriding all the highlight groups that is affected, as what I'm doing now. However, it would be better if I do not need to configure manually as I would need to refer to the source code and find the corresponding color name.

**Additional context**
\-

nullchilly:
fix #401

With `vim.o.winblend = 0` and `vim.o.pumblend = 0` (default value in neovim):

image

With `vim.o.winblend = 50` and `vim.o.pumblend` = 50:

image

### DIFF
--- a/lua/catppuccin/groups/editor.lua
+++ b/lua/catppuccin/groups/editor.lua
@@ -38,9 +38,12 @@ function M.get()
 				or C.base,
 		}, -- normal text in non-current windows
 		NormalSB = { fg = C.text, bg = C.crust }, -- normal text in non-current windows
-		NormalFloat = { fg = C.text, bg = O.transparent_background and C.none or C.mantle }, -- Normal text in floating windows.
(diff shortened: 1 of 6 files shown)


EdenEast (commented):
This looks good. Could you also just validate that this works in vim as well. From reading the code it looks like it should but I have been surprised before.

nullchilly:
Oh yeah I completely forgot that vim doesn't support `vim.o`, should we force it in?

nullchilly:
Pushed a fix for vim, seems to be working now

EdenEast (approved):

PR #411, opened 2023-02-19
feat(lspsaga): support v0.2.7

Jint-lzxy:
This PR added support for lspsaga@v0.2.7. Changes include:
- Implemented new highlight groups;
- The `LspKind*` highlight groups are used to set the colors for lspkind icons **(finally!)**
  - Note: Those highlights do _not_ belong to any integration. I added them to support "dynamically changing flavors when editing". Not sure if this is the best way to do so.

It'll be marked as ready for review after changes are merged upstream 👍


nullchilly:
In favour of directly using `SagaWinbarClass = { link = "Class }` instead of `custom_kind` because lspsaga makes no effect to preserve highlight groups on ColorScheme event

Jint-lzxy:
> In favour of directly using `SagaWinbarClass = { link = "Class }` instead of `custom_kind` because lspsaga makes no effect to preserve highlight groups on ColorScheme event

@nullchilly Sorry, I can't quite understand that. Now `lspsaga` uses highlight groups to which the `lspkind` icons are linked to set all its colors. If we only set highlights for `winbar`, the icons for other features _(such as outline)_ will have no color then. IMO `lspsaga` can handle this correctly now, see: [lspkind.lua#L110](https://github.com/glepnir/lspsaga.nvim/blob/match_theme/lua/lspsaga/lspkind.lua#L110).

Did I miss something here?

lua/catppuccin/groups/integrations/lsp_saga.lua
```diff
@@ -24,42 +25,89 @@ function M.get()
...
-		LspSagaWinbarStruct = { fg = C.mauve },
-		LspSagaWinbarEvent = { fg = C.mauve },
-		LspSagaWinbarOperator = { fg = C.green },
-		LspSagaWinbarTypeParameter = { fg = C.green },
-		LspSagaWinbarTypeAlias = { fg = C.green },
-		LspSagaWinbarParameter = { fg = C.blue },
-		LspSagaWinbarStaticMethod = { fg = C.peach },
-		LspSagaWinbarMacro = { fg = C.red },
+		-- Lspkind icons support
+		LspKindClass = { fg = C.yellow },
```
nullchilly, line 35: I mean we directly set it like this instead:

```lua
SagaWinbarClass = { link = "Class }
```

lua/catppuccin/groups/integrations/lsp_saga.lua
```diff
@@ -24,42 +25,89 @@ function M.get()
...
+		-- Microsoft-specific icons.
+		LspKindText = { fg = C.green },
+		LspKindSnippet = { fg = C.mauve },
+		LspKindFolder = { fg = C.blue },
+		LspKindUnit = { fg = C.green },
+		LspKindValue = { fg = C.peach },
+	}
+end
+
+function M.custom_kind()
```
nullchilly, line 74: So there's no need for user to manually add this to their configuration

Jint-lzxy: But if we set it this way colors are limited to `winbar` _only_, and there would be no highlight in other places that also utilizes these icons like "lspsaga outline".

nullchilly: Ah yeah sorry I overlooked the outline feature

nullchilly (approved):

lua/catppuccin/groups/integrations/lsp_saga.lua
Jint-lzxy: Maybe they can be changed to:
```lua
LspKindClass = { link = "Class" },
```
?

PR #416, opened 2023-02-21
fix need-check-nil warning

sugiura-hiromichi:
I found better way to treat *optional* value (value that could nil) and fix it in `lua/catppuccin/init.lua`.

This is my first pr, so there should be some inappropriate manner…


README.md
```diff
@@ -35,6 +35,10 @@ This port of <a href="https://github.com/catppuccin/">Catppuccin</a> is special
 <summary>Macchiato</summary>
 <img src="https://user-images.githubusercontent.com/56817415/213473285-7bd858be-6947-4d9e-8c01-2573cbc7e76c.png"/>
 </details>
+<<<<<<< HEAD
```
ToxicAven, line 38: Not sure how a merge conflict header ends up here...

ToxicAven:
Also *everything* got reformatted, pulls should probably just fix the issue instead of reformatting the entire repo.

sugiura-hiromichi:
thanks for reviewing! I fixed format & copied README.

PR #417, opened 2023-02-26
fix(compile): invalid cterm

### LINKED ISSUE #378: Setting cterm in custom_highlights causes bad compilation
### Description

Overriding cterm value in `custom_highlights` causes a wrong compilation output.

### Neovim version

```markdown
NVIM v0.8.2
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

Arch Linux

catppuccin master (3020af75aa)

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. Quit and re-run `nvim -nu minimal.lua` or run `:CatppuccinCompile`

### Expected behavior

No errors!

### Actual behavior

Loading the compiled file causes an error like this:
```
packer.nvim: Error running config for catppuccin: /tmp/catppuccin_error.lua:444: '<name>' expected near '0x7fab79d7a068'
```

And the line of the error is:
```lua
vim.api.nvim_set_hl(0, "SpellBad", { undercurl = true, cterm = table: 0x7fab79d7a068, sp = "#D20F39" })
```

### Minimal config

```Lua
-- `minimal.lua` used for reproducible configuration
-- Open with `nvim --clean -u minimal.lua`

local is_windows = vim.fn.has("win32") == 1
local function join(...)
  local sep = is_windows and "\\" or "/"
  return table.concat({ ... }, sep)
end

local root_tmp = is_windows and os.getenv("TEMP") or "/tmp"
local site_path = join(root_tmp, "nvim", "site")
local pack_path = join(site_path, "pack")
local install_path = join(pack_path, "packer", "start", "packer.nvim")
local compile_path = join(install_path, "plugin", "packer_compiled.lua")
vim.opt.packpath = site_path

local function load_plugins()
  local packer = require("packer")
  local use = packer.use
  packer.reset()
  packer.init({ compile_path = compile_path, package_root = pack_path })

  use("wbthomason/packer.nvim")
  use({ "catppuccin/nvim", as = "catppuccin" })
  -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE

  packer.install()
end

_G.load_config = function()
  -- ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
  require("catppuccin").setup({
    flavour = "mocha",
    custom_highlights = function(_)
      return {
        SpellBad = {
          cterm = { "underdashed" }
        }
      }
    end
  })

  vim.api.nvim_command("colorscheme catppuccin")
end

if vim.fn.isdirectory(install_path) == 0 then
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
vim.cmd([[autocmd User PackerComplete ++once echo "Ready!" | lua load_config()]])
```

nullchilly:
close #378

### DIFF
--- a/lua/catppuccin/lib/compiler.lua
+++ b/lua/catppuccin/lib/compiler.lua
@@ -8,8 +8,9 @@ local fmt = string.format
 local function inspect(t)
 	local list = {}
 	for k, v in pairs(t) do
-		local q = type(v) == "string" and [["]] or ""
-		table.insert(list, fmt([[%s = %s%s%s]], k, q, tostring(v), q))
+		local tv = type(v)
+		local q = tv == "string" and [["]] or ""
+		table.insert(list, fmt([[%s = %s%s%s]], k, q, tv == "table" and inspect(v) or tostring(v), q))
 	end
 	return fmt([[{ %s }]], table.concat(list, ", "))
 end
--- a/lua/catppuccin/lib/compiler.lua
+++ b/lua/catppuccin/lib/compiler.lua
@@ -9,8 +9,13 @@ local function inspect(t)
 	local list = {}
 	for k, v in pairs(t) do
 		local tv = type(v)
-		local q = tv == "string" and [["]] or ""
-		table.insert(list, fmt([[%s = %s%s%s]], k, q, tv == "table" and inspect(v) or tostring(v), q))
+		if tv ~= "table" then
+			table.insert(list, fmt([[%s = %s]], k, tostring(v)))
+		elseif tv == "string" then
+			table.insert(list, fmt([[%s = "%s"]], k, v))
+		else
+			table.insert(list, fmt([[%s = %s]], k, inspect(v)))
+		end
 	end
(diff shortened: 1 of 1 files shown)


lua/catppuccin/lib/compiler.lua
```diff
@@ -8,8 +8,9 @@ local fmt = string.format
 local function inspect(t)
 	local list = {}
 	for k, v in pairs(t) do
-		local q = type(v) == "string" and [["]] or ""
-		table.insert(list, fmt([[%s = %s%s%s]], k, q, tostring(v), q))
+		local tv = type(v)
+		local q = tv == "string" and [["]] or ""
```
EdenEast, line 12: If the type is a table then we don't need to also check if it is a string for quoting. I would have the table insert in an if block and the current impl in the else.

EdenEast (approved):

PR #427, opened 2023-03-09
feat: add support for new lsp semantic token highlight groups

mrtnvgr:
Highlighting on nightly is **broken** right now, because of neovim/neovim#22022.

This PR is basically just recent commits from folke/tokyonight.nvim.

This needs to be tested, bc I'm not sure if I didn't break anything.

### DIFF
--- a/lua/catppuccin/groups/integrations/semantic_tokens.lua
+++ b/lua/catppuccin/groups/integrations/semantic_tokens.lua
@@ -2,41 +2,6 @@ local M = {}
 
 function M.get()
 	return {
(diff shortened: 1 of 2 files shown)


nullchilly:
I have no idea why `@lsp.type.property` is linked to `Idenfifier`, doesn't seem like a good idea from upstream.

mrtnvgr:
> I have no idea why `@lsp.type.property` is linked to `Idenfifier`, doesn't seem like a good idea from upstream.

`@lsp.type.property` is linked to `@property`, isn't it?

nullchilly:
Merged as an emergency, thank you!

PR #446, opened 2023-04-02
fix(nvim-treesitter-context): make `TreesitterContext` more readable when using `transparent_background` option

### LINKED ISSUE #445: Make nvim-treesitter-context more distinguishable
When `transparent_background = true`, the floating window of nvim-treesitter-context becomes indistinguishable:

Expected behavior: 

I checked https://github.com/catppuccin/nvim/blob/73587f9c454da81679202f1668c30fea6cdafd5e/lua/catppuccin/groups/integrations/treesitter_context.lua#L6 and overridden it in my config file, it worked fine as expected:

```lua
custom_highlights = function(C)
  return {
    TreesitterContext = { bg = C.mantle, fg = C.text }
  }
end
```

The default `bg = C.none` doesn't seem to make sense.

Also, I tested [tokyonight.nvim](https://github.com/folke/tokyonight.nvim) and even though it has `floats = "transparent"` set, it is still non-transparent.

mrtnvgr:
This PR closes #445

### DIFF
--- a/lua/catppuccin/groups/integrations/treesitter_context.lua
+++ b/lua/catppuccin/groups/integrations/treesitter_context.lua
@@ -3,7 +3,6 @@ local M = {}
 function M.get()
 	return {
 		TreesitterContext = { bg = C.mantle, fg = C.text },
-		TreesitterContextLineNumber = { fg = C.surface1 },
 	}
 end
 
--- a/lua/catppuccin/groups/integrations/treesitter_context.lua
+++ b/lua/catppuccin/groups/integrations/treesitter_context.lua
@@ -2,10 +2,7 @@ local M = {}
 
 function M.get()
 	return {
-		TreesitterContext = {
-			bg = O.transparent_background and C.none or C.mantle,
-			fg = C.text,
-		},
+		TreesitterContext = { bg = C.mantle, fg = C.text },
 		TreesitterContextLineNumber = { fg = C.surface1 },
 	}
 end


89iuv (approved):

EdenEast:
Looking at the default highlight [defintions](https://github.com/nvim-treesitter/nvim-treesitter-context/blob/959265a5d438462dc5d3dde215b915109c3d6fcb/lua/treesitter-context.lua#L767-L768) do this need to be overridden? 

- By default `TreesitterContext` links to `NormalFloat`. Should it be handled like other floating windows?
- `TreesitterContextLineNumber` is the same definition as `LineNr` which it links to by default.

The integration could just be removed.

mrtnvgr:
> Looking at the default highlight [defintions](https://github.com/nvim-treesitter/nvim-treesitter-context/blob/959265a5d438462dc5d3dde215b915109c3d6fcb/lua/treesitter-context.lua#L767-L768) do this need to be overridden? 
> 
> - By default `TreesitterContext` links to `NormalFloat`. Should it be handled like other floating windows?
> - `TreesitterContextLineNumber` is the same definition as `LineNr` which it links to by default.
> 
> The integration could just be removed.

Thanks! Yes, `TreesitterContext` and `TreesitterContextLineNumber` are basically just links now. I agree that `TreesitterContextLineNumber` is obsolete here, but this PR won't fix the issue if `TreesitterContext` will be linked to `NormalFloat`.

PR #453, opened 2023-04-13
ci: add selene support

mrtnvgr:
### TODO
- [x] Add ci test
- [x] ...


.github/workflows/selene.yml
```diff
@@ -0,0 +1,21 @@
+name: Selene
+on:
+  pull_request: ~
```
nekowinston, line 3: I'd consider myself quite familiar with YAML, but had to read the docs for the use of `~` being equal to `null`.

Can we simplify this to
```suggestion
  pull_request:
```

nekowinston (commented):
Seems alright, should probably be merged as the `ci:` or `build:` scope in conventional commits.

nekowinston (approved):

PR #458, opened 2023-04-16
feat(colorschemes)!: persist g:colors_name in colorschemes

rstacruz:
Thanks so much for all the work on the Neovim theme!

Here's a slight improvement that sets the `g:colors_name` whenever you do `:colorscheme catppuccin-<variant>`. Some reasons why you'd want this:

- When using `:Telescope colorschemes` to select a variant, you can use Telescope again and the same variant will be selected.

- You might have a config that persists color schemes for the next Neovim startup. (this is what I do!)

### DIFF
--- a/lua/catppuccin/lib/compiler.lua
+++ b/lua/catppuccin/lib/compiler.lua
@@ -29,9 +29,10 @@ return string.dump(function()
 vim.o.termguicolors = true
 if vim.g.colors_name then vim.cmd("hi clear") end
 vim.o.background = "%s"
-vim.g.colors_name = "catppuccin"
+vim.g.colors_name = "catppuccin-%s"
 local h = vim.api.nvim_set_hl]],
-			flavour == "latte" and "light" or "dark"
+			flavour == "latte" and "light" or "dark",
+			flavour
 		),
 	}
 	if path_sep == "\\" then O.compile_path = O.compile_path:gsub("/", "\\") end
--- a/lua/catppuccin/lib/vim/compiler.lua
+++ b/lua/catppuccin/lib/vim/compiler.lua
@@ -17,8 +17,9 @@ if exists("colors_name")
 endif
 set termguicolors
 set background=%s
-let g:colors_name = "catppuccin"]],
-			(flavour == "latte" and "light" or "dark")
+let g:colors_name = "catppuccin-%s"]],
+			(flavour == "latte" and "light" or "dark"),
+			flavour
 		),
 	}
 


mrtnvgr:
Thanks for your contribution. I think the best way to implement this is changing `g:colors_name` in compiler functions.

```diff
diff --git a/lua/catppuccin/lib/compiler.lua b/lua/catppuccin/lib/compiler.lua
index f4d32fd..a0ced28 100644
--- a/lua/catppuccin/lib/compiler.lua
+++ b/lua/catppuccin/lib/compiler.lua
@@ -29,9 +29,10 @@ return string.dump(function()
 vim.o.termguicolors = true
 if vim.g.colors_name then vim.cmd("hi clear") end
 vim.o.background = "%s"
-vim.g.colors_name = "catppuccin"
+vim.g.colors_name = "catppuccin-%s"
 local h = vim.api.nvim_set_hl]],
-                       flavour == "latte" and "light" or "dark"
+                       flavour == "latte" and "light" or "dark",
+                       flavour
                ),
        }
        if path_sep == "\\" then O.compile_path = O.compile_path:gsub("/", "\\") end
diff --git a/lua/catppuccin/lib/vim/compiler.lua b/lua/catppuccin/lib/vim/compiler.lua
index 6ca202e..7cd3593 100644
--- a/lua/catppuccin/lib/vim/compiler.lua
+++ b/lua/catppuccin/lib/vim/compiler.lua
@@ -17,8 +17,9 @@ if exists("colors_name")
 endif
 set termguicolors
 set background=%s
-let g:colors_name = "catppuccin"]],
-                       (flavour == "latte" and "light" or "dark")
+let g:colors_name = "catppuccin-%s"]],
+                       (flavour == "latte" and "light" or "dark"),
+                       flavour
                ),
        }

```

mrtnvgr:
This is a breaking change, so it should be marked with an exclamation mark.

rstacruz:
Thanks! I'm not too familiar with the code at the moment to do that right now - I assume using `catppuccin-%s` will need to be changed to account for the "default" flavour (just `catppuccin`)?

mrtnvgr:
> Thanks! I'm not too familiar with the code at the moment to do that right now 

If you want, I can do that for you.

> I assume using `catppuccin-%s` will need to be changed to account for the "default" flavour (just `catppuccin`)?

Sorry, I didn't understand your question. Could you write it in other words please?

rstacruz:
> Sorry, I didn't quite understand your question. Could you write it in other words please?

In the case of doing `:colorscheme catppuccin`, I was wondering if this line:

```
+vim.g.colors_name = "catppuccin-%s"
```

might set `"catppuccin-"` (with a trailing `-`) instead of "catppuccin".

mrtnvgr:
> > Sorry, I didn't quite understand your question. Could you write it in other words please?
> 
> In the case of doing `:colorscheme catppuccin` (no flavour), I was wondering if this line:
> 
> ```
> +vim.g.colors_name = "catppuccin-%s"
> ```
> 
> might set `"catppuccin-"` (with a trailing `-`) instead of "catppuccin".

It will be set to your default flavour from configuration.

```
:colorscheme catppuccin
:lua print(vim.g.colors_name) -> "catppuccin-mocha"
```

rstacruz:
I've updated the PR with the patch above. Thanks for the assist!

PR #461, opened 2023-04-18
fix(gitsigns): respect transparency

emxxjnm:
respect transparency for word diff in preview

### DIFF
--- a/lua/catppuccin/groups/integrations/gitsigns.lua
+++ b/lua/catppuccin/groups/integrations/gitsigns.lua
@@ -1,34 +1,49 @@
 local M = {}
 
 function M.get()
-	return {
-		GitSignsAdd = { fg = C.green }, -- diff mode: Added line |diff.txt|
-		GitSignsChange = { fg = C.yellow }, -- diff mode: Changed line |diff.txt|
-		GitSignsDelete = { fg = C.red }, -- diff mode: Deleted line |diff.txt|
+	if type(O.integrations.gitsigns) == "boolean" then
+		O.integrations.gitsigns = { enabled = true, transparent = false }
+	end
 
-		GitSignsAddPreview = O.transparent_background and { fg = U.darken(C.green, 0.72, C.base), bg = C.none }
-			or { link = "DiffAdd" },
-		GitSignsDeletePreview = O.transparent_background and { fg = U.darken(C.red, 0.72, C.base), bg = C.none }
-			or { link = "DiffDelete" },
-		-- for word diff in previews
-		GitSignsAddInline = O.transparent_background and {
-			fg = C.green,
-			bg = C.none,
-			style = { "bold" },
-		} or { link = "DiffAdd" },
-		GitSignsDeleteInline = O.transparent_background and {
-			fg = C.red,
-			bg = C.none,
-			style = { "bold" },
-		} or { link = "DiffDelete" },
-		GitSignsChangeInline = O.transparent_background and {
-			fg = C.yellow,
-			bg = C.none,
-			style = { "bold" },
-		} or { link = "DiffChange" },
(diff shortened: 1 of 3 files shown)


lua/catppuccin/groups/integrations/gitsigns.lua
```diff
@@ -6,8 +6,28 @@ function M.get()
 		GitSignsChange = { fg = C.yellow }, -- diff mode: Changed line |diff.txt|
 		GitSignsDelete = { fg = C.red }, -- diff mode: Deleted line |diff.txt|
 
-		GitSignsAddPreview = O.transparent_background and { fg = C.green, bg = C.none } or { link = "DiffAdd" },
-		GitSignsDeletePreview = O.transparent_background and { fg = C.red, bg = C.none } or { link = "DiffDelete" },
+		GitSignsAddPreview = O.transparent_background and { fg = U.darken(C.green, 0.72, C.base), bg = C.none }
+			or { link = "DiffAdd" },
```
mrtnvgr, line 10:
```suggestion
		GitSignsAddPreview = O.transparent_background and { link = "diffAdded" } or { link = "DiffAdd" },
```

lua/catppuccin/groups/integrations/gitsigns.lua
```diff
@@ -6,8 +6,28 @@ function M.get()
...
+		GitSignsDeletePreview = O.transparent_background and { fg = U.darken(C.red, 0.72, C.base), bg = C.none }
+			or { link = "DiffDelete" },
```
mrtnvgr, line 12:
```suggestion
		GitSignsDeletePreview = O.transparent_background and { link = "diffRemoved" } or { link = "DiffDelete" },
```

emxxjnm:
befor:
WeChat8e99bec8d779e421d4ba7d7e765ca44c

after:
WeChatb9e19851fc8ee02db071c23d1048f0ef


Darken Add/DeletePreview, it seems to be able to better highlight word diff 😺

nekowinston:
While this seems like a nice change word changes in the line, I'm not sold on complete line changes being dimmed. In your preview, the fact that overlay was changed to subtext isn't as visible to me as in the before screenshot.

2023-04-20-054000@2x

I'd like to know what other people think of this change, but I think this is a case where visibility > appearance.

mokrinsky:
I don't really like this change.
1) *befor* catches attention of the user much faster than *after*. In case of editing large files it's important.
2) There are enough fonts where bold looks almost the same as normal weight, so *after* will become mostly useless.

mrtnvgr:
> 2. There are enough fonts where bold looks almost the same as normal weight, so _after_ will become mostly useless.

The difference between these screenshots is not only in the boldness of the text, but also in the brightness.  The modified text is lighter than the rest of the line.

emxxjnm:
> While this seems like a nice change word changes in the line, I'm not sold on complete line changes being dimmed. In your preview, the fact that overlay was changed to subtext isn't as visible to me as in the before screenshot.
> 
> 2023-04-20-054000@2x
> 
> I'd like to know what other people think of this change, but I think this is a case where visibility > appearance.

This is a controversial change, in the premise of transparent background, some people like to remove the background color, through the light and dark to show the difference, so need to see other people's opinion or if there is a better solution to remove the background color

sxyazi:
I prefer the before one for its recognizability over beauty when diffing.😸

mokrinsky:
We talked about it a bit in discord, I thought someone will offer it here.

Maybe it would be better to define an optional parameter. Smth like:
```lua
require("catppuccin").setup {
    integrations = {
        gitsigns = {
            enabled = true,
            transparent = true,
        },
    },
}
```

PR #462, opened 2023-04-19
fix(barbecue): apply same bg as statusline has

ArtAndreev:
I suggest to set the same background color for winbar (barbecue), as statusline (lualine) has, for consistency:

image

I looked at VSCode, status line has the same color as tab line has and context line is slightly brighter than both, but darker than regular text background. Maybe the color of winbar or statusline should be changed to another color or adjusted to VSCode colors.

### DIFF
--- a/lua/barbecue/theme/catppuccin.lua
+++ b/lua/barbecue/theme/catppuccin.lua
@@ -5,8 +5,10 @@ local dirname_color = O.integrations.barbecue.dim_dirname and C.overlay1 or C.te
 local basename_bold = O.integrations.barbecue.bold_basename
 local context_color = O.integrations.barbecue.dim_context and C.overlay1 or C.text
 
+local transparent_bg = O.transparent_background and "NONE" or C.mantle
+
 local M = {
-	normal = { fg = C.text, bg = "none" },
+	normal = { fg = C.text, bg = transparent_bg },
 
 	ellipsis = { fg = C.overlay1 },
 	separator = { fg = C.overlay1 },


mrtnvgr:
Thanks!

thanhvule0310:
I feel this change is inappropriate, Most popular themes (including all default themes) of Vscode have background color that should sync with the code background color instead of the status bar. If you really want to have consistency between them then just overwrite in setup (But now this commit has been merged and the current one to overwrite is me.)

mrtnvgr:
> I feel this change is inappropriate, Most popular themes (including all default themes) of Vscode have background color that should sync with the code background color instead of the status bar. If you really want to have consistency between them then just overwrite in setup (But now this commit has been merged and the current one to overwrite is me.)

Yes, you are right. I'm sorry. [Even other neovim colorschemes set their backgrounds like that...](https://github.com/folke/tokyonight.nvim/blob/ff81eac0ecf85db235dfc4c9d54f2d07d662a423/lua/barbecue/theme/tokyonight.lua#L4) I think we can make a new setting in barbecue integration config. I will make a PR ASAP.

thanhvule0310:
> > I feel this change is inappropriate, Most popular themes (including all default themes) of Vscode have background color that should sync with the code background color instead of the status bar. If you really want to have consistency between them then just overwrite in setup (But now this commit has been merged and the current one to overwrite is me.)
> 
> Yes, you are right. I'm sorry. [Even other neovim colorschemes set their backgrounds like that...](https://github.com/folke/tokyonight.nvim/blob/ff81eac0ecf85db235dfc4c9d54f2d07d662a423/lua/barbecue/theme/tokyonight.lua#L4) I think we can make a new setting in barbecue integration config. I will make a PR ASAP.

Well, thank you for taking your time to maintain this great project.

ArtAndreev:
Sorry for being annoying.
I think, referencing another themes is not a good way for making decisions – they might made a mistake and barbecue was created recently, so little time has passed and maintainers haven't decided yet about colors.

Now the winbar and the code (editor section) have the same background color, so it's difficult to separate them visually. I wrote in PR description, that colors might should be adjusted.

Screenshot 2023-04-20 at 09 18 27

Look at cattpuccin behavior in VSCode:

Screenshot 2023-04-20 at 09 04 33

The theme looks more consistent and has a better visual separation between sections. Also barbecue (winbar) was inspired by VSCode: [README.md](https://github.com/utilyre/barbecue.nvim/blob/main/README.md?plain=1#L5).

I think, catppuccin should have strong defaults and in this situation too.

PR #471, opened 2023-04-28
feat(alpha): add support for it

hmajid2301:
Hi :wave:,

I am using [alpha-vim](https://github.com/goolord/alpha-nvim)  extension. Specifically the way [LazyVim](https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/ui.lua#L267-L275) has configured it

 I wanted to add the catppuccin theme to it, else it was all white.

This is what it looks like: 

image

Open to feedback thanks!

### DIFF
--- a/vim.yml
+++ b/vim.yml
@@ -209,6 +209,9 @@ globals:
   O.integrations.dashboard:
     type: bool
     property: read-only
+  O.integrations.alpha:
+    type: bool
+    property: read-only
   O.integrations.gitsigns:
     type: bool
     property: read-only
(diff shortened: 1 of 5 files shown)


vim.yml
```diff
@@ -209,6 +209,9 @@ globals:
   O.integrations.dashboard:
     type: bool
     property: read-only
+  O.integrations.alpha:
```
hmajid2301, line 212: wasn't sure the best place to put this

mrtnvgr: > wasn't sure the best place to put this

1. alpha
1. cmp
1. dashboard

nullchilly:
pink is too personal for AlphaButtons, I think if we are going to enable alpha integration by default it should be something simply, for example:

image

PR #484, opened 2023-05-09
fix: make `CursorLine` visible for `transparent_background`

loqusion:
Resolves a conflict between `catppuccin`'s transparent mode and `neo-tree` where `neo-tree`'s fuzzy search would have an invisible `CursorLine`, making it hard to know which item is being selected.

## Before

image

## After

image


mrtnvgr (approved):
Thanks!

nullchilly:
If anyone want to disable cursorline in transparent mode, please use `:set cursorlineopt=number`

PR #502, opened 2023-06-03
fix typo wrt fidget.nvim setting

tsnieman:
fixes typo wrt fidget.nvim setting — it looks like a copy-paste-o from the row above.


nekowinston (approved):
Thank you!

PR #506, opened 2023-06-08
docs: fix lazy.nvim install method

emivespa:
as per the docs, colorschemes should be eager-loaded loaded with higher than default priority

> Only useful for start plugins (lazy=false) to force loading certain plugins first. Default priority is 50. It's recommended to set this to a high number for colorschemes.
- https://github.com/folke/lazy.nvim#-plugin-spec

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ This port of <a href="https://github.com/catppuccin/">Catppuccin</a> is special 
 
 [lazy.nvim](https://github.com/folke/lazy.nvim)
 ```lua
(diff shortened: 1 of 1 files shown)


mrtnvgr (approved):

nullchilly:
https://github.com/folke/lazy.nvim#-colorschemes

No need to add lazy = false spec

And `priority = 1000` is the convention here

emivespa:
does `priority` have any effect when lazy-loading?

PR #509, opened 2023-06-12
fix(feline): use new `vim.lsp.status()` method

### LINKED ISSUE #508: Feline not updated to nvim 0.10
### Description

feline integration not working on nvim 0.10

this is due to `vim.lsp.util.get_progress_messages` being deprecated [see here](https://github.com/neovim/neovim/blob/a7e5d4238a00d5bfa5809b2860047eca2d565e62/runtime/doc/news.txt#L151-L152)

`vim.lsp.status` should be used instead

### Neovim version

```markdown
NVIM v0.10.0-dev-499+g643546b82
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Terminal and multiplexer

kitty 0.28.1

head

### Steps to reproduce

update neovim to lastest nightly release
run it

### Expected behavior

just works

### Actual behavior

Error message
image

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "catppuccin/nvim",
  "famiu/feline.nvim"
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here

local ctp_feline = require('catppuccin.groups.integrations.feline')

ctp_feline.setup{}

require("feline").setup({
    components = ctp_feline.get(),
})
```

mrtnvgr:
Fixes #508

### DIFF
--- a/lua/catppuccin/groups/integrations/feline.lua
+++ b/lua/catppuccin/groups/integrations/feline.lua
@@ -293,7 +293,12 @@ function M.get()
 	components.active[2][1] = {
 		provider = function()
 			-- local Lsp = vim.lsp.status()[1]
-			local Lsp = vim.fn.has "neovim-0.10" == 1 and vim.lsp.util.get_progress_messages()[1] or vim.lsp.status()[1]
+			local Lsp
+			if vim.fn.has "neovim-0.10" then
+				Lsp = vim.lsp.util.get_progress_messages()[1]
+			else
+				Lsp = vim.lsp.status()[1]
+			end
 
 			if Lsp then
 				local msg = Lsp.message or ""
--- a/lua/catppuccin/groups/integrations/feline.lua
+++ b/lua/catppuccin/groups/integrations/feline.lua
@@ -292,7 +292,8 @@ function M.get()
 	-- workspace loader
 	components.active[2][1] = {
 		provider = function()
-			local Lsp = vim.lsp.util.get_progress_messages()[1]
+			-- local Lsp = vim.lsp.status()[1]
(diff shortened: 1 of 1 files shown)


lua/catppuccin/groups/integrations/feline.lua
```diff
@@ -292,7 +292,8 @@ function M.get()
 	-- workspace loader
 	components.active[2][1] = {
 		provider = function()
-			local Lsp = vim.lsp.util.get_progress_messages()[1]
+			-- local Lsp = vim.lsp.status()[1]
+			local Lsp = vim.fn.has "neovim-0.10" == 1 and vim.lsp.util.get_progress_messages()[1] or vim.lsp.status()[1]
```
lucario387, line 296: Uhh, this returns 0, so this breaks for neovim-stable users

nekowinston: Yeah, this errors for me on 0.9.1, this works:
```suggestion
			local Lsp
			if vim.fn.has "neovim-0.10" then
				Lsp = vim.lsp.util.get_progress_messages()[1]
			else
				Lsp = vim.lsp.status()[1]
			end
```

lua/catppuccin/groups/integrations/feline.lua
```diff
@@ -292,7 +292,12 @@ function M.get()
 	-- workspace loader
 	components.active[2][1] = {
 		provider = function()
-			local Lsp = vim.lsp.util.get_progress_messages()[1]
+			local Lsp
+			if vim.fn.has "neovim-0.10" then
```
lucario387, line 296: This is still wrong. It should be `vim.fn.has "nvim-0.10" == 1` as 0 is considered true in lua

PR #510, opened 2023-06-13
fix(feline): disable lsp status on nightly

mrtnvgr:
Hotfix for #509

### DIFF
--- a/lua/catppuccin/groups/integrations/feline.lua
+++ b/lua/catppuccin/groups/integrations/feline.lua
@@ -292,12 +292,7 @@ function M.get()
 	-- workspace loader
 	components.active[2][1] = {
 		provider = function()
-			local Lsp
-			if vim.fn.has "nvim-0.10" == 0 then
-				Lsp = vim.lsp.util.get_progress_messages()[1]
-			else
-				Lsp = vim.lsp.status()[1]
-			end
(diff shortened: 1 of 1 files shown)


nekowinston (approved):
Sorry for overlooking this as well, I was only testing on stable.

nullchilly:
Please don't use version check

nullchilly:
Also I just tested and `vim.lsp.status()[1]` doesn't work, did you test the PR at all before mering?

It should be `vim.lsp.status()` instead.

mrtnvgr:
> Also I just tested and `vim.lsp.status()[1]` doesn't work, did you test the PR at all before merging?
> 
> It should be `vim.lsp.status()` instead.

I couldn't test myself, so I looked for similar fixes from other repositories. I saw suggestions and assumed that others had already checked it. I'm sorry.

nullchilly:
Seems like it still doesn't work since the new `vim.lsp.status()` returns `87%: Loading workspace: 82/101, Loading workspace: 83/94` instead of a `{ message, percentage, tittle }`

nullchilly:
The new output returns `73%: Loading workspace: 69/101, Loading workspace: 69/94` which breaks when feline do a string.format on it, i.e `string.format("%:")`

I think we should just disable this component on nightly instead of fighting to workaround.

PR #512, opened 2023-06-18
feat(semantic_tokens): add some lsp semantic tokens

yavorski:
Add few/missing LSP Semantic Tokens: 

* `@lsp.type.boolean`
* `@lsp.type.builtinType`
* `@lsp.type.number`
* `@lsp.type.typeAlias`
* `@lsp.type.unresolvedReference`
* `@lsp.typemod.class.defaultLibrary`

### DIFF
--- a/lua/catppuccin/groups/integrations/semantic_tokens.lua
+++ b/lua/catppuccin/groups/integrations/semantic_tokens.lua
@@ -12,6 +12,8 @@ If you want to stay on nvim 0.7, disable the integration.
 		return {}
 	end
 	return {
+		["@lsp.type.boolean"] = { link = "@boolean" },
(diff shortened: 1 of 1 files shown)


mrtnvgr (approved):

nullchilly:
I just hope that there's saner upstream default for semantic tokens, i.e. https://github.com/neovim/neovim/pull/22981 so we don't need to play the catching game like this.

@mrtnvgr FWIW you can usually double check these type of PR with [tokyonight's semantic tokens](https://github.com/folke/tokyonight.nvim/blob/main/lua/tokyonight/theme.lua#L267)

One of the charm of catppuccin is we won't leave anyone behind (issues or pull requests) so don't feel the need to wait for another approval as long as you tested the PR before merging.

PR #521, opened 2023-06-24
feat(telescope): telescope flat style support

nottyl:
Updates the Telescope integration to allow users to choose a telescope style between "nvchad" and "classic". 
The default style currently is set to "classic"

- Classic (the original style):

Screenshot 2023-06-24 at 11 46 38

- Nvchad (flat style)

Screenshot 2023-06-24 at 11 47 07

### DIFF
--- a/doc/catppuccin.txt
+++ b/doc/catppuccin.txt
@@ -706,7 +706,7 @@ telekasten.nvim>lua
 telescope.nvim>lua
     telescope = {
 	enabled = true,
-	style = "classic" | "flat",
+	style = "classic" | "nvchad",
     }
 <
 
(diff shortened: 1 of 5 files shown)


nullchilly:
Please keep backwards compatibility, `O.integrations.telescope` was a boolean before and can't be indexed

The default should be classic since we respect plugin's theme guideline

I think we should rename `flat -> nvchad` since it's the original

image

nottyl:
Ok! I changed the name `flat` to `nvchad` and included backward compatibility in the latest commit. Though I don't know how to implement backward compatibility correctly so the code might be a bit ugly :<

mrtnvgr (approved):
LGTM! Thanks!

PR #526, opened 2023-06-24
fix(native_lsp): boolean logic

### LINKED ISSUE #523: bg error(from prev commit)
### Description

Error starting from `b032ced` commit. My config https://github.com/geril2207/nvim-config
E5113: Error while calling lua chunk: vim/_editor.lua:0: /Users/ilya/.config/nvim/init.lua..nvim_exec2() called at /Users/ilya/.config/nvim/init.lua:0../Users/ilya/.local/share/nvim/lazy/catppuccin/colo
rs/catppuccin.vim, line 1: Vim(lua):E5108: Error executing lua /Users/ilya/.cache/nvim/catppuccin/macchiato:0: Invalid 'bg': expected String or Integer
stack traceback:
        [C]: in function 'nvim_set_hl'
        /Users/ilya/.cache/nvim/catppuccin/macchiato: in function 'f'
        ...local/share/nvim/lazy/catppuccin/lua/catppuccin/init.lua:137: in function 'load'

### Neovim version

```markdown
NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3
```

### Terminal and multiplexer

iterm2

last

### Steps to reproduce

Check my config https://github.com/geril2207/nvim-config, config theme in /after/plugin/theme.lua

### Expected behavior

Should work

### Actual behavior

Throws errors

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  "catppuccin/nvim",
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

Yuhf7:
Fixes #523 
The error seems to be due to boolean logic in `native_lsp.lua`:
```diff
-			bg = O.transparent_background and C.none
+			bg = O.transparent_background
+				or not inlay_hints.background and C.none
```

Since `and` has priority over `or`, Lua computes `not inlay_hints.background and C.none` first, and then `O.transparent_background` or'd with the result of `and`. 
Then, if `O.transparent_background` is true (which is the case for me), the `bg` field is set to true, which causes the "invalid 'bg'" error.

Surrounding the `or` statement with parens seemed to solve the issue for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,7 +409,7 @@ coc_nvim = true,
 
 In the inners tables you can set the style for the diagnostics, both `virtual_text` (what you see on the side) and `underlines` (what points directly at the thing (e.g. an error)).
 
-```
+```lua
 native_lsp = {
     enabled = true,
     virtual_text = {
@@ -423,6 +423,9 @@ native_lsp = {
         hints = { "underline" },
         warnings = { "underline" },
         information = { "underline" },
+    },
+    inlay_hints = {
+        background = true,
     },
 },
 ```
@@ -930,6 +933,9 @@ native_lsp = {
         hints = { "underline" },
         warnings = { "underline" },
         information = { "underline" },
+    },
+    inlay_hints = {
+        background = true,
     },
 },
 ```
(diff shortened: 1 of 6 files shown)


mrtnvgr (approved):
Thanks!

PR #529, opened 2023-07-01
fix(mapper): remove unnecessary globals

LordMZTE:
This uses Lua's environment features in order to modify global scope for loading hlgroups instead of the old hacky solution which modified `_G`.


nullchilly:
This PR is creating side effect of random flavour during startup, still no idea why.

Issue #525, opened 2023-06-24
Share your catppuccin config to prevent bug regressions


nullchilly:
Configs in this thread will be added in [user-configs branch](https://github.com/catppuccin/nvim/blob/user-configs/tests/setup_spec.lua) and [tested before merging any PR](https://github.com/catppuccin/nvim/actions/runs/5510065469/jobs/10043707281#step:7:62) to prevent bug regressions.

You can either post your config table or link to your config!

nullchilly:
The initial step is done here https://github.com/catppuccin/nvim/blob/12894370fa3c9e2200f3724c4184354d6b79733a/tests/setup_spec.lua#L14-L96

Will need to ask the community for more configs

nullchilly:
```
{ 
 	{ 
 		term_colors = true, 
 		transparent_background = false, 
 		styles = { 
 			comments = {}, 
 			conditionals = {}, 
 			loops = {}, 
 			functions = {}, 
 			keywords = {}, 
 			strings = {}, 
 			variables = {}, 
 			numbers = {}, 
 			booleans = {}, 
 			properties = {}, 
 			types = {}, 
 		}, 
 		color_overrides = { 
 			mocha = { 
 				base = "#000000", 
 				mantle = "#000000", 
 				crust = "#000000", 
 			}, 
 		}, 
 		integrations = { 
 			dropbar = { 
 				enabled = true, 
 				color_mode = true, 
 			}, 
 		}, 
 	}, 
 }
 ```

**Comment by mrtnvgr** [comment]
I guess mine is the smallest here 😅
https://github.com/mrtnvgr/init.lua/blob/main/lua/initlua/plugins/colorschemes/pre/catppuccin.lua

**Comment by Rishabh672003** [comment]
heres mine - https://github.com/Rishabh672003/Neovim/blob/main/lua/rj/plugins/catppuccin.lua#L10-L44

**Comment by adelarsq** [comment]
Here is mine. Just relevant parts:

```lua
-- Using Lazy
{ 'https://github.com/catppuccin/nvim', event = 'VeryLazy', name = 'catppuccin', },

...

local hour = tonumber(fn.strftime('%H'))
if hour >= 17 or hour < 7 then
    o.background = 'dark'
else
    o.background = 'light'
end

if o.background == 'light' then
    cmd.colorscheme 'catppuccin-latte'
else
    cmd.colorscheme 'catppuccin-frappe'
end
```

LordMZTE:
[...]

PR #532, opened 2023-07-07
test: move setup spec to user-configs branch

nullchilly:
Related: https://github.com/catppuccin/nvim/issues/525

Perhaps we should move this chunk of test to an orphan branch?


nekowinston:
I'm all for using a orphan branch for this.

We can keep the file in `tests/setup_spec.lua`, then run tests on `main`, use `git checkout user-configs tests/setup_spec.lua` before we actually run the tests. As long as we don't commit back onto the main branch in the action, this will keep the branches apart cleanly.

@mrtnvgr `test:` is a type in Conventional Commits, changing this to `ci(tests):` wasn't necessary.

nekowinston (approved):
Turns out it wasn't as easy as simply running `git checkout user-configs tests/setup_spec.lua` in GitHub actions, but we got there in the end :slightly_smiling_face:

PR #537, opened 2023-07-10
docs: update types

nekowinston:
Adds more explicit docs to the `types.lua` file, and adds all the integrations that were missing from the `integrations` table.

### DIFF
--- a/lua/catppuccin/types.lua
+++ b/lua/catppuccin/types.lua
@@ -3,97 +3,246 @@
 ---@field setup fun(opts: CatppuccinOptions?)
 
 ---@alias CtpFlavor "mocha" | "macchiato" | "frappe" | "latte"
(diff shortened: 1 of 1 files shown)


lua/catppuccin/types.lua
```diff
@@ -3,97 +3,246 @@
...
+-- Handles the style of general hl groups (see `:h highlight-groups`).
+---@field styles CtpStyles
+-- Toggle integrations. Integrations allow Catppuccin to set the theme of various plugins.
+---@field integrations CtpIntegrations
+-- Catppuccin colors can be overwritten here.
+---@field color_overrides CtpColors | CtpFlavors<CtpColors<string>>
+-- Catppuccin highlights can be overwritten here.
+---@field highlight_overrides CtpHighlightOverrides
+-- deprecated: use highlight_overrides instead.
+---@field custom_highlights CtpHighlightOverrideFn
```
nekowinston, line 39: @nullchilly should this be marked as deprecated?
There are still [configs](https://github.com/catppuccin/nvim/blob/7a9437f54ae9d96105dfe41958892e7287adf5ac/tests/setup_spec.lua#L526-L529) using it in our tests, and this annotation alone will not affect the plugin, but might keep people from using it in newer setups.

nekowinston: So passing in a table
```lua
custom_highlights = {
  TelescopeTitle = "#ff69b4",
}
```
should also show as correctly typed, gotcha.

PR #540, opened 2023-07-10
fix(telescope): keep consistency between the two styles

Jint-lzxy:
Just borrowed several highlights from the original integration :)


nullchilly (approved):

nullchilly:
Thanks! The flamingo caret sure looks nice :)

PR #542, opened 2023-07-11
fix(telescope): respect transparency

emxxjnm:
before:
image

after:
image


lua/catppuccin/groups/integrations/telescope.lua
```diff
@@ -32,11 +32,11 @@ function M.get()
 				bg = O.transparent_background and C.none or C.red,
 			},
 			TelescopeResultsTitle = {
-				fg = C.mantle,
+				fg = O.transparent_background and C.peach or C.mantle,
```
nullchilly, line 35: I think it should be the same color as no transparent background AKA `C.lavender`

PR #553, opened 2023-07-19
chore: update pull request template

nekowinston:
Changes wording for our PR template:
- `types.lua` is added, so we don't have to play catch up with new integrations again.
- Mention that our vimdoc is auto-generated, so people don't change it manually
- Since 9d562c161ef6febbef1d04eef111882a176c651e we're using release-please, so we should enforce Conventional Commits rather than just recommending them.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,6 +5,7 @@ Here are some guidelines:
 - New plugin integration should be added in alphabetical order:
   - to the [README](https://github.com/catppuccin/nvim#integrations) (vimdoc is auto-generated).
   - to [types.lua](https://github.com/catppuccin/nvim/blob/main/lua/catppuccin/types.lua)
+  - to [vim.yml](https://github.com/catppuccin/nvim/blob/main/vim.yml)
```
nekowinston, line 8: Thanks for adding this

PR #559, opened 2023-07-28
fix(integration_default): override `enabled` key only if integration has one

mrtnvgr:
Fixes #558

PR #562, opened 2023-07-29
refactor(init): integration_default

mrtnvgr:


nullchilly:
`integration_default` is causing way too much maintenance burden:

- #559 Implemented inside setup function, doesn't leverage #534

- #560 This is a monkeypatch used only for bufferline because of a past mistake. I do not encourage patching the `.get()` function.

- #562 `integration_default = true` is currently useless because all are already enabled by default (Except navic which should have been true by default)

- Currently, the only use for this is to disable all integrations by default

With the above points, I think it's best we remove this feature for now because the current solution is suboptimal.

PR #564, opened 2023-08-05
feat(treesitter-context): add color for normal background

ofseed:
The only effect of the former treesitter-context support is to add a border-bottom when the `transparent_background` is on.

This PR adds two highlights for normal background, one for a looming border-bottom, and one for the line number.

Former:

image

Current:

image

NOTE: The `TreesitterContextLineNumber` should not be configured in catppuccin because the nvim-treesitter-context provided an option to decide whether accept this highlight.

### DIFF
--- a/lua/catppuccin/groups/integrations/treesitter_context.lua
+++ b/lua/catppuccin/groups/integrations/treesitter_context.lua
@@ -1,7 +1,9 @@
 local M = {}
 
 function M.get()
-	return O.transparent_background and { TreesitterContextBottom = { sp = C.dim, style = { "underline" } } } or {}
+	return O.transparent_background and {
+		TreesitterContextBottom = { sp = C.dim, style = { "underline" } },
+	} or {}
 end
 
 return M
--- a/lua/catppuccin/groups/integrations/treesitter_context.lua
+++ b/lua/catppuccin/groups/integrations/treesitter_context.lua
@@ -3,7 +3,12 @@ local M = {}
 function M.get()
 	return O.transparent_background and {
 		TreesitterContextBottom = { sp = C.dim, style = { "underline" } },
-	} or {}
+	} or {
+		TreesitterContextLineNumber = {
+			fg = C.surface1,
+			bg = C.mantle,
+		},
+	}
 end
 
 return M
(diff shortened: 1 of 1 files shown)


lua/catppuccin/groups/integrations/treesitter_context.lua
```diff
@@ -1,7 +1,18 @@
 local M = {}
 
 function M.get()
-	return O.transparent_background and { TreesitterContextBottom = { sp = C.dim, style = { "underline" } } } or {}
+	return O.transparent_background and {
+		TreesitterContextBottom = { sp = C.dim, style = { "underline" } },
+	} or {
+		TreesitterContextBottom = {
+			sp = C.surface0,
```
nullchilly, line 8: Shouldn't this be fg instead of sp?

# With sp

image

# With fg

image

ofseed: You may be using WezTerm. It seems that WezTerm does not support the underline of having another color for the word. In the terminal I used, Kitty(what I'm using), Alacritty, Gnome Terminal, and Konsole all support this feature but WezTerm does not.

nekowinston: image
I'm using the latest nightly build of WezTerm, I think this looks as it should? @ofseed

nekowinston: I'm sorry I misread the PR, I thought that was *for* transparency.
Here's the PR with transparency disabled, in WezTerm:
image

nullchilly: Quoting from [wezterm faq](https://wezfurlong.org/wezterm/faq.html#how-do-i-enable-undercurl-curly-underlines):

Note: on Windows, the ConPTY layer strips out the curly underline escape sequences. If you're missing this feature in your WSL instance, you will need to use either `wezterm ssh` or [multiplexing](https://wezfurlong.org/wezterm/multiplexing.html#connecting-into-windows-subsystem-for-linux) to bypass ConPTY.

nekowinston: I think we can mark this as resolved, then?

nekowinston (approved):

PR #566, opened 2023-08-16
Add `nvim-window-picker` integration

mehalter:
This adds integration for [nvim-window-picker](https://github.com/s1n7ax/nvim-window-picker). It is used by `neo-tree` and they recently added the ability to configure the colors through highlight groups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1094,7 +1094,7 @@ ts_rainbow = true
 <td>
 
 ```lua
-window_picker = true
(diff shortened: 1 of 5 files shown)


mehalter:
Let me know if there is a standard color that is used in other integrations to make the picker visible and noticeable. Basically these highlight groups indicate something to be selected. I made them red so that they jump out to the user

PR #579, opened 2023-09-02
feat: add workaround for kitty transparent issue

### LINKED ISSUE #578: Disable Transparent background doesn't work if kitty share the same background
Kitty's opacity setting will change background if they share the same color:

> Note that it only sets the background color’s opacity in cells that have the same background color as the default terminal background,

So if I use the same flavour in nvim and kitty, and set kitty transparent, then nvim will be transparent whether `transparent_background` is true or false

nullchilly:
Close #578


nekowinston (approved):

opalmay:
Now I have the opposite issue - when background is different it'll always be transparent

PR #605, opened 2023-11-01
fix(flash): link `FlashPrompt` to `NormalFloat`

### LINKED ISSUE #604: Flash prompt appears black when winblend is set
### Description

If `winblend` is set, flash prompt's background will appear black.

### Neovim version

```markdown
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1697887905
```

### Terminal and multiplexer

alacritty 0.12.3 (5efb0690)

main / b359f3e

### Steps to reproduce

1. `nvim -u repro.lua`
2. Press `s` to open flash's jump
3. Notice the prompt is black

### Expected behavior

It should behave like other float windows.

### Actual behavior

The prompt's background is black instead.

image

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
	{
		"catppuccin/nvim",
		name = "catppuccin",
		lazy = true,
		init = function()
			vim.opt.winblend = 10 -- important
		end,
		opts = {
		},
	},
	-- add any other plugins here
	{
		"folke/flash.nvim",
		event = "VeryLazy",
		opts = {},
		keys = {
			{
				"s",
				mode = { "n", "x", "o" },
				function()
					require("flash").jump()
				end,
				desc = "Flash",
			},
		},
	},
}
	root = root .. "/plugins",
})

-- add anything else here
```

jamestansx:
Fixes #604

### DIFF
--- a/lua/catppuccin/groups/integrations/flash.lua
+++ b/lua/catppuccin/groups/integrations/flash.lua
@@ -7,6 +7,7 @@ function M.get()
 		FlashLabel = { fg = C.green, bg = bg, style = { "bold" } },
 		FlashMatch = { fg = C.lavender, bg = bg },
 		FlashCurrent = { fg = C.peach, bg = bg },
+		FlashPrompt = { bg = (O.transparent_background and vim.o.winblend == 0) and C.none or C.mantle },
 	}
 end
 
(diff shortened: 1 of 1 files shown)


mrtnvgr:
Thanks for your contribution! <3

PR #615, opened 2023-11-16
docs(fidget.nvim): `window.blend` -> `notification.window.winblend`

jdekarske:
Fixed special setup according to https://github.com/j-hui/fidget.nvim/blob/main/doc/fidget.md#options


mrtnvgr:
Thanks!

PR #618, opened 2023-11-26
feat(neotree): make popup titlebar text visible

portus-zephyr:
makes popup titlebar text more visible when neotree `popup_border_style` = `NC`


nullchilly:
Can confirm 👍🏻 

image

PR #624, opened 2023-12-07
refactor(dap): split dap and dap_ui integrations

igorlfs:
This Pull Request makes 3 changes to the nvim-dap integration:
1. Splits the dap and dap_ui integrations. Rationale: there's no need for both plugins to share a integration. In fact, that's the only integration that contains multiple plugins.
2. Adds a missing highlight group to the nvim-dap integration
3. Removes a misguided comment about enabling the nvim-dap integration

I'm not sure why, but the integrations are not enabled by default. Did I forget something?


nullchilly:
1. Need to add backwards compatibility because we're no longer 0ver
```lua
dap = { enabled = true, dap_ui = true }
```
should be auto converted to
```lua
dap = true, dap_ui = true
```
Something like this would work (Please verify:
```lua 
if O.integrations.dap.enabled and O.integrations.dap.dap_ui then
  O.integrations.dap_ui = O.integrations.dap.dap_ui
end
```
2. Integrations are enabled by default if they're known to be well tested.

nullchilly:
Failed on `require("catppuccin").setup { integrations = { dap = true } }`

```lua
Failed to run `config` for catppuccin

...chilly/code/git/catppuccin/lua/catppuccin/lib/mapper.lua:30: attempt to index field
'dap' (a boolean value)
```

PR #630, opened 2023-12-18
feat(treesitter): follow upstream captures

ofseed:
nvim-treesitter may be updated soon, still in discussion but it is planned, see https://github.com/nvim-treesitter/nvim-treesitter/issues/4799#issuecomment-1860154411. The purpose is to follow the treesitter common core to share queries, so some captures will be redefined.

I created this PR to follow upstream changes to avoid highlights break by the breaking change.

There are some highlights defined in treesitter common core but not mentioned in nvim-treesitter

- `@carriage-return`
- `@comment.documentation`
- `@embedded`
- `@escape`
- `@markup.link`
- `@markup.quote`
- `@property.builtin`
- `@variable.member`

And not sure all helix captures will be applied. This PR only follow the common core, need to continue to follow the implementation.

### DIFF
--- a/lua/catppuccin/groups/integrations/treesitter.lua
+++ b/lua/catppuccin/groups/integrations/treesitter.lua
@@ -13,21 +13,18 @@ If you want to stay on nvim 0.7, either disable the integration or pin catppucci
 	end
 
 	local colors = { -- Reference: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md
+		-- Identifiers
+		["@variable"] = { fg = C.text, style = O.styles.variables or {} }, -- Any variable name that does not have another highlight.
+		["@variable.builtin"] = { fg = C.red, style = O.styles.properties or {} }, -- Variable names that are defined by the languages, like this or self.
+		["@variable.parameter"] = { fg = C.maroon, style = O.styles.variables or {} }, -- For parameters of a function.
+		["@variable.member"] = { fg = C.lavender }, -- For fields.
 
-		-- Misc
-		["@error"] = { link = "Error" },
-		["@operator"] = { link = "Operator" }, -- For any operator: +, but also -> and * in C.
+		["@constant"] = { link = "Constant" }, -- For constants
+		["@constant.builtin"] = { fg = C.peach, style = O.styles.keywords or {} }, -- For constant that are built in the language: nil in Lua.
+		["@constant.macro"] = { link = "Macro" }, -- For constants that are defined by macros: NULL in C.
 
-		-- Comment
-		["@comment"] = { link = "Comment" },
-		["comment.note"] = { fg = C.base, bg = C.blue },
-		["comment.warning"] = { fg = C.base, bg = C.yellow },
-		["comment.error"] = { fg = C.base, bg = C.red },
-
-		-- Punctuation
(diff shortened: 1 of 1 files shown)


lucario387:
Hello, nvim-treesitter maintainer here 👋 I believe it's better to link new hlgroups to old ones for the time being, i.e. `["@string.regexp"] = { link = "@string.regex"}, ["@string.regex"] = {...}` to keep compatibility with both current and soon-default branch of nvim-treesitter

wookayin:
For the legacy highlights, instead of setting a separate highlight, `highlight link` would be a better idea.

ofseed:
I think If the upstream does not do this, the downstream should not add it either, and for the legacy branch users, the highlights be linked to mean nothing because they should not exist yet.

For me recording a macro to assign again is simpler than linking them. :D

However, I will consider it carefully again before the PR gets merged.

ofseed:
There are two extra highlight groups added:

- `@markup.list`, to reduce abuse of @punctuation.special
- `@diff.delta`, even not used in queries, I link it to `diffChanged`  like the convention

One highlights group was split:

- `@text.uri` is now `@markup.link.uri` in markup links and `@string.special.url` outside of markup.   I assigned the value of `@text.uri` to `@markup.link.url`.

lucario387:
Updates here :)
`@text.reference -> @markup.link` and `@field -> @variable.member`, `@markup.quote` is now also added

ofseed:
The refactor is merged into nvim-treesitter experimental branch `main`. I followed it up and it is ready to merge now.

lucario387:
nvim-treesitter/nvim-treesitter#5895 is the PR that will backport the changes to `master`. So you can test it there.

ofseed:
LGTM.

lua/catppuccin/groups/integrations/treesitter.lua
```diff
@@ -12,122 +12,131 @@ If you want to stay on nvim 0.7, either disable the integration or pin catppucci
...
+		-- Comment
+		["@comment"] = { link = "Comment" },
 
-		["@storageclass"] = { link = "StorageClass" }, -- visibility/life-time/etc. modifiers (e.g. `static`)
-		["@attribute"] = { link = "Constant" }, -- attribute annotations (e.g. Python decorators)
-		["@field"] = { fg = C.lavender }, -- For fields.
-		["@property"] = { fg = C.lavender, style = O.styles.properties or {} }, -- Same as TSField.
+		["comment.error"] = { fg = C.base, bg = C.red },
+		["comment.warning"] = { fg = C.base, bg = C.yellow },
+		["comment.note"] = { fg = C.base, bg = C.blue },
```
lucario387, line 92: missing `comment.todo`. Also they should be `@comment.*`

ofseed: The former `@text.todo` has two different meanings, 
- one for todo in comments, now `@comment.warning`
- one for todo lists, now `@markup.list.checked`/`@markup.list.unchecked`

So the `comment.todo` is a new capture, I think it's better not to add new highlights for now, especially since this highlight hasn't been utilized.(it's not exists in nvim-treesitter now)

nullchilly:
Thank you so much for the PR!

lua/catppuccin/groups/integrations/treesitter.lua
ofseed: Continue on https://github.com/catppuccin/nvim/pull/645

PR #645, opened 2024-01-19
feat: use a more distinguishable color for todos

ofseed:
Previously, there was no distinction between the colors of `@comment.todo` and `@comment.warn.` The reason is that the color of the warning and the color of the todo are both yellow. I suggest changing the color of the todo to a slightly different color.

image


lucario387:
Also we renamed `comment.note` to `comment.hint` in nvim-treesitter/nvim-treesitter#5895 as well. The queries are being updated to reflect that.

Issue #657, opened 2024-02-13
Wrong color shown when using `color_overrides` in Kitty


jerboa88:
### Description

When using kitty terminal, certain colors added to the `color_overrides` config section are shown incorrectly. For example, setting `text = "#dedeff"` results in yellow text being shown.

Cause: The [workaround for kitty transparency implementation](https://github.com/catppuccin/nvim/pull/579) in lua/catppuccin/palettes/init.lua treats the input color as a single integer, and incrementing this integer can cause the green and blue channels to roll over to 0 if they are already at max brightness.

### Neovim version

```markdown
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1692716794
```

### Terminal and multiplexer

kitty 0.32.1

### Catppuccin version / branch / rev

main

### Steps to reproduce

1. Run `nvim -u repro.lua` in kitty terminal
2. Observe the text color is yellow, rather than white

### Expected behavior

Displayed colors should match what is defined in `color_overrides`

### Actual behavior

Displayed colors are incorrect when defining colors in `color_overrides` that end in `ff` or `ffff`

### Repro

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
	vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
	{
		"catppuccin/nvim",
		name = "catppuccin",
		opts = {
			color_overrides = {
				all = {
					text = "#dedeff", -- This will be displayed incorrectly as #dede00
					-- text = "#dedefe", -- This will be displayed correctly as #dedeff
				}
			}
		},
	},
}
require("lazy").setup(plugins, {
	root = root .. "/plugins",
})

vim.cmd.colorscheme("catppuccin")
-- add anything else here
```

PR #658, opened 2024-02-13
fix: wrong color shown when using color_overrides

jerboa88:
Fix for #657 (Wrong color shown when using color_overrides in Kitty).

Rewrote this section of code to increment or decrement just the blue channel instead of treating color as a single number.

### DIFF
--- a/lua/catppuccin/palettes/init.lua
+++ b/lua/catppuccin/palettes/init.lua
@@ -7,9 +7,12 @@ function M.get_palette(flavour)
 	local ans = vim.tbl_deep_extend("keep", O.color_overrides.all or {}, O.color_overrides[flvr] or {}, palette or {})
 	if O.kitty then -- https://github.com/kovidgoyal/kitty/issues/2917
 		for accent, hex in pairs(ans) do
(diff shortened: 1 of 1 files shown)


mrtnvgr:
This does not work for me

20240213_09h43m27s_grim

jerboa88:
@mrtnvgr I think lazy is still using a cached version of catppuccin/nvim in your case.

After running repro.lua from #657, I had to delete the .repro directory so that lazy.nvim will pull it again from my repo. After that, running `nvim -u repro.lua` with your repro.lua works as expected.

nullchilly (approved):
I use [github cli](https://cli.github.com/) and testing this PR is as simple as `gh pr checkout 658` inside catppuccin git folder

<table>
<tr>
<td></td> <td> #dedefe </td> <td> #dedeff </td>
</tr>
<tr>

<td>

# Before

</td>

<td>

image

</td>
<td>

image

</td>
</tr>

<tr>

<td>

# After

</td>

<td>

image

</td>
<td>

image

</td>
</tr>

</table>

Thanks for the PR!

PR #659, opened 2024-02-14
feat(treesitter): add styles.miscs to disable hardcoded italics

GotaLoveFiraCode:
Set everything to `O.styles.conditionals or { "italic" }`, so, by default, the various settings are still italic. In other words, this is not a breaking change.

🎉 First off, thanks for taking the time to contribute! 🎉

Here are some guidelines:
- Format code using [stylua](https://github.com/johnnymorganz/stylua).
- Create a topic branch on your fork for your specific PR.
- Use [conventionalcommits.org's](https://www.conventionalcommits.org/en/v1.0.0/)
  rules for explicit and meaningful commit messages.
- If it's your first time contributing to a project, then read
  [About pull requests](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests)
  on Github's docs.

Here are some tips:
- Use `vim.g.catppuccin_debug = true` to get live config re-loading

### DIFF
--- a/lua/catppuccin/groups/integrations/treesitter.lua
+++ b/lua/catppuccin/groups/integrations/treesitter.lua
@@ -23,7 +23,9 @@ If you want to stay on nvim 0.7, either disable the integration or pin catppucci
 		["@constant.builtin"] = { fg = C.peach, style = O.styles.keywords or {} }, -- For constant that are built in the language: nil in Lua.
 		["@constant.macro"] = { link = "Macro" }, -- For constants that are defined by macros: NULL in C.
 
-		["@module"] = { fg = C.lavender, style = { "italic" } }, -- For identifiers referring to modules and namespaces.
+		-- This is difficult to avoid hardcoding, as there is no user option for this (yet).
+		-- For now, binding to `conditionals` because it is italic by default.
+		["@module"] = { fg = C.lavender, style = O.styles.conditionals or { "italic" } }, -- For identifiers referring to modules and namespaces.
 		["@label"] = { link = "Label" }, -- For labels: label: in C and :label: in Lua.
 
 		-- Literals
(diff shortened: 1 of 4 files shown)


mrtnvgr:
Catppuccin compiles your configuration to simple colors, it helps to reduce start up time. `no_italic` (as well as other `no_...` options) removes all italic from highlight groups on compile time (https://github.com/catppuccin/nvim/blob/b76ada82bf2019d5e343018b4104cc9266900c16/lua/catppuccin/lib/compiler.lua#L52). If you still have italics, we should debug [that](https://github.com/catppuccin/nvim/blob/b76ada82bf2019d5e343018b4104cc9266900c16/lua/catppuccin/lib/compiler.lua#L52) code.

lua/catppuccin/groups/integrations/treesitter.lua
```diff
@@ -23,7 +23,9 @@ If you want to stay on nvim 0.7, either disable the integration or pin catppucci
 		["@constant.builtin"] = { fg = C.peach, style = O.styles.keywords or {} }, -- For constant that are built in the language: nil in Lua.
 		["@constant.macro"] = { link = "Macro" }, -- For constants that are defined by macros: NULL in C.
 
-		["@module"] = { fg = C.lavender, style = { "italic" } }, -- For identifiers referring to modules and namespaces.
+		-- This is difficult to avoid hardcoding, as there is no user option for this (yet).
+		-- For now, binding to `conditionals` because it is italic by default.
+		["@module"] = { fg = C.lavender, style = O.styles.conditionals or { "italic" } }, -- For identifiers referring to modules and namespaces.
```
lucario387, line 28: This isn't a conditional though?. You can always just `style = {}` in userconfig or is it not

GotaLoveFiraCode:
> Catppuccin compiles your configuration to simple colors, it helps to reduce start up time. `no_italic` (as well as other `no_...` options) removes all italic from highlight groups on compile time. If you still have italics, we should debug [that](https://github.com/catppuccin/nvim/blob/b76ada82bf2019d5e343018b4104cc9266900c16/lua/catppuccin/lib/compiler.lua#L52) code.

Hello! Sorry for taking so long to reply, I’ve been traveling.
When setting `no_italic`, everything works as expected.
What I am trying to fix with this PR is that it is impossible to control the italics that catppuccin’s treesitter integration sets. An example is anything treesitter thinks of as a module will become italic (e.g. in `mod somemodule` in rust, the `somemodule` would be italic), even if the user doesn’t want it to be italic. Currently the only option is to have all modules (among other things) be italic, or turn italics off completely with `no_italic`—there is no way for the user to control modules specifically. Ideally, there should be a setting in `styles` in the setup call, so the user can control, independently from catppuccin’s treesitter integration, if modules should be italic or not. For now though, setting modules to the users `conditionals` setting is easier and keeps the same default.

lua/catppuccin/groups/integrations/treesitter.lua
GotaLoveFiraCode: This precisely is the problem; the user’s config does not override this setting. That’s what I changed, by binding the setting for `@module` to the users setting for conditionals (which, by default, is also italic). This way the user can change if modules are italic or not, while the default of italic is still preserved.

nullchilly:
You can now do

```lua
require("catppuccin").setup { styles = { miscs = {} } }
```

To disable the remaining hardcoded styles

PR #685, opened 2024-03-26
feat(feline): fix feline integration

89iuv:
This PR is intended to fix the feline integration by:
- changing the background color from mantle to surface0 which is the actual color used in the pictures
- enable the user to use the setup function without any parameter as described in the documentation

Visual changes:
Old:
image

New:
image

Reference:

### DIFF
--- a/lua/catppuccin/groups/integrations/feline.lua
+++ b/lua/catppuccin/groups/integrations/feline.lua
@@ -26,7 +26,7 @@ local assets = {
 
 local sett = {
 	text = C.mantle,
-	bkg = C.mantle,
+	bkg = C.surface0,
 	diffs = C.mauve,
 	extras = C.overlay1,
 	curr_file = C.maroon,
@@ -71,7 +71,11 @@ function M.setup(opts)
 		opts.sett = opts.sett or {}
 		opts.mode_colors = opts.mode_colors or {}
 	else
-		opts = {}
+		opts = {
+			assets = {},
+			sett = {},
+			mode_colors = {},
+		}
 	end
 	assets = vim.tbl_deep_extend("force", assets, opts.assets)
 	sett = vim.tbl_deep_extend("force", sett, opts.sett)


sgoudham:
We're using mantle bars across our editors now afaik so I believe the picture should be updated instead of reverting back to surface0

mrtnvgr (approved):

lua/catppuccin/groups/integrations/feline.lua
```diff
@@ -26,7 +26,7 @@ local assets = {
 
 local sett = {
 	text = C.mantle,
-	bkg = C.mantle,
+	bkg = C.surface0,
```
sgoudham, line 29:
```suggestion
	bkg = C.mantle,
```

lua/catppuccin/groups/integrations/feline.lua
```diff
@@ -71,7 +71,11 @@ function M.setup(opts)
 		opts.sett = opts.sett or {}
 		opts.mode_colors = opts.mode_colors or {}
 	else
-		opts = {}
+		opts = {
+			assets = {},
+			sett = {},
```
sgoudham, line 76: I think this should be `set` ?

sgoudham (changes requested):
As mentioned above, the neovim screenshot should be updated to have the crust bar

89iuv:
Lookin at the VScode repo i see that we are using Crust for the status bar, not mantle.

The problem that i am trying to fix is that in nvim the status bar and the file explorer have the same background color making them blend on into the other and not providing a clear visual distinction between the two of them.

I think if we want to keep consistency across catppuccin editors then maybe crust would be a better fit:
image

sgoudham:
Ah yes, good catch! Crust sounds good for consistency with VSCode. @89iuv

lua/catppuccin/groups/integrations/feline.lua
89iuv: In the current implementation:
if you use the documentation to create the feline config:
```
local ctp_feline = require('catppuccin.groups.integrations.feline')

ctp_feline.setup()
```

What will happen is: 

- a new empty table will be created on line 74: ```opts = {}```.
- after that on line 80: ```assets = vim.tbl_deep_extend("force", assets, opts.assets)``` this error will be thrown: "vim/shared.lua:0: after the second argument: expected table, got nil" because the opt.assets is nil. 

This change eliminates this issue;

Note: 

- this does not happen if you pass an empty table to ctp_feline so updating the documentation would have the same effect.

sgoudham: Sorry, didn't realise the variable is meant to be `sett`

Thanks for the info

sgoudham (approved):
Thanks for updating the colours, don't want to block the merge on screenshots so we can update them at a later time 👍

PR #688, opened 2024-03-27
feat(feline): improve feline lsp display

89iuv:
This PR adds the option to changes the "Lsp" text to the actual name of the connected lsp servers.

It also adds the option to disable the lsp progress indicator. 
The reason for this is that there are other solutions to display the progress indicator which work better.

Old:
image

New:
image
image


mrtnvgr (approved):
Nice!

PR #691, opened 2024-03-28
Refresh airline colors

MrJulianBlack:
Hello, first off, this is my first PR ever, so sorry in advance if not everything is up to standard, I tried to follow every information i could find.

The purpose of this PR is to fix some errors with the Airline integration. I know there is also support for Lualine, but i haven't personally made that switch yet.

### VCS section styling:

For some reason, section B of Airline (second from the left, where i.e. Git branch is displayed) is styled exactly the same before as section C (Filename). By changing the background color, it looks much better. This is the same color that is used in Lualine for this purpose.

### Replace mode:

I don't know if anyone uses replace mode not out of accident, but I noticed the colors for warning & error section were not applied in that mode. I changed this to match all the other modes by defining replace mode as a variation of insert mode. **Different than in Lualine,** I only set the red color on the left side of the status-line, because it is already used on the right for the error-section and looks kind of wrong otherwise.

I also could not find any documentation what so ever on Airline for "insert_replace", so assuming it is a deprecated configuration option, I removed it.

### Terminal+Command mode:

I added different colors for these modes, where before they only used the color from normal mode. Terminal mode is defined as an exact copy of insert mode, which is the same configuration as used for Lualine. For command mode, I defined it as a variation of normal mode where the orange color (same as in Lualine) is only appiled on the left side, for the same reason as for replace mode, it is already used on the right side.

So please, let me know what you think! Obviously, colors are a subjective topic, but I feel like at least the changes for the VCS section are needed and where unintended before. For everything else, as I said, i tried to be as close to Lualine in everything to achieve the same same style. If needed, I can take some time to take before&after screenshots showing the changes I made.

### DIFF
--- a/autoload/airline/themes/catppuccin.vim
+++ b/autoload/airline/themes/catppuccin.vim
@@ -20,7 +20,8 @@ function! airline#themes#catppuccin#refresh()
 	let s:V3 = [ s:c.text, s:c.mantle, 145, 16	] " guifg guibg ctermfg ctermbg
 
 	" Replace mode
-	let s:RE = [ s:c.mantle, s:c.red, 59, 203 ] " guifg guibg ctermfg ctermbg
+	let s:R1 = [ s:c.mantle, s:c.red, 59, 203 ] " guifg guibg ctermfg ctermbg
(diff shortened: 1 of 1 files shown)


nullchilly:
Thanks!

# Before

image

# After

image

PR #718, opened 2024-06-10
fix(defaults): Enable mini integration by default

### LINKED ISSUE #715: Add support for `mini.statusline`
Currently this plugin does not support the highlight groups used by the plugin `mini.statusline`

Implementing the catppuccin theme colors for the highlight groups defined [here](https://github.com/echasnovski/mini.statusline/blob/94d5e48415bdf872536e5812475fcf19e09f5c0e/lua/mini/statusline.lua#L49).

tjkuson:
The documentation says that mini is enabled by default, but it is not. This patch changes the defaults to match what is documented.

https://github.com/catppuccin/nvim/blob/cc8e290d4c0d572171243087f8541e49be2c8764/README.md?plain=1#L880-L885

Closes #715

### DIFF
--- a/lua/catppuccin/init.lua
+++ b/lua/catppuccin/init.lua
@@ -103,6 +103,10 @@ local M = {
 				enabled = false,
 				color = "red",
 			},
+			mini = {
+				enabled = true,
+				indentscope_color = "text",
+			},
 		},
 		color_overrides = {},
 		highlight_overrides = {},


vollowx:
Also set a indentscope color like surface2

tjkuson:
> Maybe also set a indentscope color like `surface2` ?

Good point, I forgot to actually set one. I have set the indentscope colour to `text` as that's what the current document suggests. Do you think it would be better if I set it to `surface2` and update the documentation as well?

sgoudham:
Hey :wave: 
Thanks for catching this.

I don't think you _need_ to set `indentscope_color` since we seem to default to `text`:
https://github.com/catppuccin/nvim/blob/67565cd353fa543fa30cb738570c2e4c87da3e9c/lua/catppuccin/groups/integrations/mini.lua#L16

(But I do like the explicitness anyways!)

sgoudham (approved):

PR #720, opened 2024-06-20
feat(csv): built-in rainbow highlighting

igorlfs:
Hello!

[Recently](https://github.com/neovim/neovim/pull/29395), neovim added support for CSV files. Here's how it looks:

current csv support

Which is fine, but I took the opportunity to define the groups in a rainbow-like manner, which I believe to be nicer:

new, rainbow-like

In total, there are 9 groups. The colors are inspired by the `rainbowX` groups, but since there are 3 extra colors, I've split _sapphire_ into _sky_ and _blue_ and added _mauve_ and _pink_.


vollowx (approved):
LGTM!

PR #166, opened 2022-07-05
Add explicit support for 'mini.nvim'

echasnovski:
Hi! I would like to add explicit support for [mini.nvim](https://github.com/echasnovski/mini.nvim).

Basic choices are taken from either highlight groups to which 'mini.nvim' makes default links or from analogous plugins.

Differences from default linked groups:
- 'mini.indentscope' is based on 'indent_blankline.lua'.
- 'mini.jump' is based on 'vim_sneak.lua'.
- 'mini.jump2d' is based on 'hop.lua'.
- 'mini.starter' is based on 'dashboard.lua', 'which_key.lua', and personal choices:
    - `MiniStarterInactive` is same as comments.
    - `MiniStarterItemBullet` is blue as borders.
    - `MiniStarterQuery` is green to be opposite of pink item's prefix.
    - `MiniStarterSection` is flamingo to be visible.
- 'mini.statusline' is based on 'lualine/themes/catppuccin.lua' and personal choices:
    - `MiniStatuslineDevinfo` and `MiniStatuslineFileinfo` are "slightly different text".
    - `MiniStatuslineModeOther` has teal background so that be separable from other modes.
- 'mini.tabline' is based on 'bufferline.lua':
    - `MiniTablineVisible` is "slightly different text".
    - `MiniTablineModified*` groups have inverted `fg` and `bg` of their counterparts.
- 'mini.test' has red for fail and green for pass.
- 'mini.trailspace' has group with red background to draw attention.

Before:

https://user-images.githubusercontent.com/24854248/177376092-2e9836aa-c845-4781-b79a-fbdf4f3175a8.mp4

After:

https://user-images.githubusercontent.com/24854248/177376123-c61ea18e-39b8-4a42-95b1-8811297f2b35.mp4

PR #721, opened 2024-06-24
feat(mini): add new highlight groups

echasnovski:
Add explicit support for all new groups since #166 and for the (hopefully) upcoming 'mini.icons'.

### DIFF
--- a/lua/catppuccin/groups/integrations/mini.lua
+++ b/lua/catppuccin/groups/integrations/mini.lua
@@ -8,17 +8,99 @@ function M.get()
 
 	local indentscope_color = O.integrations.mini.indentscope_color
 	return {
+		MiniAnimateCursor = { style = { "reverse", "nocombine" } },
+		MiniAnimateNormalFloat = { link = "NormalFloat" },
+
+		MiniClueBorder = { link = "FloatBorder" },
+		MiniClueDescGroup = { link = "DiagnosticFloatingWarn" },
+		MiniClueDescSingle = { link = "NormalFloat" },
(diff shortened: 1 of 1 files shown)


vollowx (approved):

lua/catppuccin/groups/integrations/mini.lua
```diff
@@ -8,17 +8,99 @@ function M.get()
...
+		MiniDepsPlaceholder = { link = "Comment" },
+		MiniDepsTitle = { link = "Title" },
+		MiniDepsTitleError = { bg = C.red, fg = C.base },
+		MiniDepsTitleSame = { link = "DiffText" },
+		MiniDepsTitleUpdate = { bg = C.green, fg = C.base },
+
+		MiniDiffSignAdd = { fg = C.green },
+		MiniDiffSignChange = { fg = C.yellow },
+		MiniDiffSignDelete = { fg = C.red },
+		MiniDiffOverAdd = { link = "DiffAdd" },
```
vollowx, line 41: Just wondering: shouldn't these linked highlights be linked inside `mini.nvim`? So that there only need to be the necessary ones that must be manually colored to look correct.

echasnovski: All of these links *are* done in 'mini.nvim', but those are default links: they take effect only if color scheme doesn't define the corresponding highlight groups.

Not having those explicit links will work if this is the only loaded color scheme. But it might not fully work if `:colorscheme catppuccin` is executed *after* applying color scheme that overrides some highlight groups.

vollowx: Got it

PR #725, opened 2024-06-28
feat(feline): add lazy.nvim updates module, replace deprecated API

ilias777:
Add option to show how many plugin updates has lazy.nvim in feline.

The code already checks if lazy.nvim is installed. If so, then the updates appears as a number with the package icon.

It looks like this:

feline-plugin-updates

### DIFF
--- a/lua/catppuccin/groups/integrations/feline.lua
+++ b/lua/catppuccin/groups/integrations/feline.lua
@@ -293,7 +293,7 @@ function M.get()
 	-- macro
 	components.active[1][12] = {
 		provider = "macro",
-		enabled = function() return vim.api.nvim_get_option "cmdheight" == 0 end,
+		enabled = function() return vim.api.nvim_get_option_value("cmdheight", { scope = "global" }) == 0 end,
 		hl = {
 			fg = sett.extras,
 			bg = sett.bkg,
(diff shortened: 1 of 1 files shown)


lua/catppuccin/groups/integrations/feline.lua
```diff
@@ -305,6 +305,24 @@ function M.get()
 		},
 		left_sep = invi_sep,
 	}
+
+	-- lazy.nvim plugin updates
+	components.active[1][14] = {
+		provider = require("lazy.status").updates,
+		enabled = function()
+			local lazypath = vim.fn.isdirectory(vim.fn.stdpath "data" .. "/lazy/lazy.nvim")
```
vollowx, line 313: Not every user uses this path, IMO using `pcall(require, 'lazy')` to check might be a more generic solution.

vollowx (changes requested):

lua/catppuccin/groups/integrations/feline.lua
```diff
@@ -305,6 +305,24 @@ function M.get()
 		},
 		left_sep = invi_sep,
 	}
+
+	-- lazy.nvim updates
+	components.active[1][14] = {
+		provider = require("lazy.status").updates,
```
vollowx, line 311: I guess this line is calling lazy.nvim no matter the `enabled` function returns true or false, once this file is loaded.

As you can see below, the 'workspace loader' module, the `provider` is also a function so that it won't call something that doesn't exist.

vollowx (changes requested):
I'm not very familiar with feline.nvim (haven't tried yet actually) but I guess this is a common behaviour of Lua, feel free to point out if I'm wrong.

lua/catppuccin/groups/integrations/feline.lua
ilias777: Ok the provider is also done with a protected call. It works as expected.

Also I update the deprecated api functions with the last commit.

vollowx:
Looks nice :heart:

PR #727, opened 2024-07-03
fix(dapui): border match bg color

igorlfs:
Simple change following 31fcfb0

Before:
image
After:
image


vollowx (approved):

vollowx:
First of all, thanks a lot for your contribution. After receiving some negative UI changes on plugins like `toggleterm.nvim` and `telescope.nvim` from users, I think it'll be better to apply #722 after more [considerations](https://github.com/catppuccin/nvim/commit/31fcfb02c47952d5c75aec893b93b2878abe5fbb#commitcomment-143916118), and the related PRs are going to be closed.

igorlfs:
Regardless of reverting #722, this is an overall improvement if we ever change the `FloatBorder` group.

PR #728, opened 2024-07-06
fix(leap): highlight group for `LeapLabel`

PlatyPew:
Changes in https://github.com/ggandor/leap.nvim/commit/b75a86f14ebdb3940460bfd1a02224210eccc698 causes leap to not highlight labels found by leap.

This PR fixes it by removing `LeapLabelSecondary` and replacing `LeapLabelPrimary` with `LeapLabel`


vollowx (approved):
LGTM!

PR #729, opened 2024-07-09
feat(Gitsigns): Darken highlight on staged changes

lasse16:
Hi,
Many thanks for the colorscheme!

I changed the highlight groups for staged changes into darker version of the unstaged changes.

## Previously it looked like this:

unstaged, following GitSignsAdd/GitSignsChange/GitSignsDelete
grafik

staged, no clue what highlight group this linked to, prob DiffAdded or something like this
grafik

## Now it looks like this:
unstaged:
grafik
staged: following GitSignsStagedAdd/GitSignsStagedChange/GitSignsStagedDelete
grafik

As you can see, the sign gutter color just darkens.

This relies on the undocumented highlight groups of GitSignsStaged...
They are not mentioned in the README and do not have a help tag.

There are also more, for in-line highlighting and number column highlighting but I dont use those features so did not include the corresponding highlight groups.

## Bonus
This is the code snippet I used to get the same effect without this PR being merged.
```lua
			local utils = require 'catppuccin.utils.colors'
			return {
				GitSignsStagedAdd = { fg = utils.darken(colors.green, 0.5, colors.base) },
				GitSignsStagedChange = { fg = utils.darken(colors.yellow, 0.5, colors.base) },
				GitSignsStagedDelete = { fg = utils.darken(colors.red, 0.5, colors.base) }
			}
		end,
```

### DIFF
--- a/lua/catppuccin/groups/integrations/gitsigns.lua
+++ b/lua/catppuccin/groups/integrations/gitsigns.lua
@@ -11,6 +11,10 @@ function M.get()
 			GitSignsChange = { fg = C.yellow }, -- diff mode: Changed line |diff.txt|
 			GitSignsDelete = { fg = C.red }, -- diff mode: Deleted line |diff.txt|
 
+			GitSignsStagedAdd = { fg = U.darken(C.green, 0.5, C.base) },
+			GitSignsStagedChange = { fg = U.darken(C.yellow, 0.5, C.base) },
+			GitSignsStagedDelete = { fg = U.darken(C.red, 0.5, C.base) }
+
 			GitSignsCurrentLineBlame = { fg = C.surface1 },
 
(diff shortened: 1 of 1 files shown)


vollowx:
Hi there :hand: . First of all, thanks for your contribution. As I can see with my configuration (which doesn't modify anything related to gitsigns.nvim), the appearance is just like the one you wanted:

|Staged|Unstaged|
|---|---|
|image|image|

And actually, this dimmed effect probably is done internally by gitsigns.nvim at [here](https://github.com/lewis6991/gitsigns.nvim/blob/375c44bdfdde25585466a966f00c2e291db74f2d/lua/gitsigns/highlight.lua#L141). Please make sure that you've set up catppuccin.nvim and enabled gitsigns.nvim integration correctly, and have the two plugins up-to-day.

PR #731, opened 2024-07-10
feat(feline): allow to hide lazy.nvim updates

vollowx:
This PR added an option to hide the count of updatable plugins from lazy.nvim, and updated the documentations for feline.nvim.

Please check this out @ilias777 :eyes:

### DIFF
--- a/lua/catppuccin/groups/integrations/feline.lua
+++ b/lua/catppuccin/groups/integrations/feline.lua
@@ -32,7 +32,7 @@ local sett = {
 	curr_file = C.maroon,
 	curr_dir = C.flamingo,
 	show_modified = false,
-	show_lazy_updates = true,
+	show_lazy_updates = false,
 }
 
 if require("catppuccin").flavour == "latte" then
--- a/lua/catppuccin/groups/integrations/feline.lua
+++ b/lua/catppuccin/groups/integrations/feline.lua
@@ -32,6 +32,7 @@ local sett = {
 	curr_file = C.maroon,
 	curr_dir = C.flamingo,
 	show_modified = false,
+	show_lazy_updates = true,
 }
 
 if require("catppuccin").flavour == "latte" then
@@ -137,12 +138,6 @@ function M.get()
 			end
 		end
 		return false
-	end
-
-	-- Check if lazy.nvim is installed
-	local has_lazy = function()
-		local lazy_installed, _ = pcall(require, "lazy")
-		return lazy_installed
 	end
 
 	-- #################### STATUSLINE ->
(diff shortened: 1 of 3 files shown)


ilias777:
Really good and very clean. I like that. 👍🏻

But you still want the `pcall`. I saw you have remove the `has_lazy()` function and at line number 314 is checking for that function.

ilias777:
Or you can set the default value to `false` and only if the users want this feature, they can enable it.

Keep in mind that this feature in lazy.nvim options (that is the `checker` table) is by default set to `false`. Only if this is set to `true`, we can use it in feline with this settings.

lua/catppuccin/groups/integrations/feline.lua
```diff
@@ -314,15 +309,9 @@ function M.get()
...
 	-- lazy.nvim updates
 	components.active[1][14] = {
-		provider = function()
-			if has_lazy() then
-				return require("lazy.status").updates()
-			else
-				return " "
-			end
-		end,
+		provider = function() return require("lazy.status").updates() end,
```
ilias777, line 312: I think here a function is not necessarily, if we don't check for any condition.

And then it looks like this: `provider = return require("lazy.status").updates` without the parenthesis. But I will check.

ilias777 (changes requested):
If we want to check for lazy.nvim installation, we need the `has_lazy()` function.

Otherwise the provider has to be:
`provider = require("lazy.status").updates`
without the parenthesis, like I mentioned above, but I can check it only if an update is available from a plugin.

ilias777:
I tested the suggested settings. I remove only the `has_lazy()` at line 314 and all works fine. 

I like it 👍🏻

lua/catppuccin/groups/integrations/feline.lua
vollowx: If the requiring isn't wrapped by a function, it will be called once the file is loaded no matter the `enabled` returns true or false, which means it might cause an error if the user doesn't use lazy.nvim

vollowx:
So I will add more info to the docs, and I guess it's fine to merge after that?

vollowx:
@ilias777 It should work normally now.

lua/catppuccin/groups/integrations/feline.lua
ilias777: I understand. Than it's better to be a function. 👍🏻

ilias777:
Yes, it works normally now. I testet it with the `pcall` in the if condition. Really good!

🙂👍🏻

PR #735, opened 2024-07-23
feat: add `grug-far.nvim` integration

mehalter:
This adds support for [grug-far.nvim](https://github.com/MagicDuck/grug-far.nvim). Most of the highlights are solid choices out of the box, but this one should be made more visible.

### DIFF
--- a/doc/catppuccin.txt
+++ b/doc/catppuccin.txt
@@ -542,6 +542,10 @@ gitsigns.nvim>lua
     gitsigns = true
 <
 
+grug-far.nvim>lua
+    grug_far = false
+<
+
 harpoon>lua
     harpoon = false
 <
--- a/README.md
+++ b/README.md
@@ -707,6 +707,17 @@ gitsigns = true
 ```
 <!-- gitsigns.nvim -->
 
+<!-- grug-far.nvim -->
+</tr>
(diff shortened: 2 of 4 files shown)


doc/catppuccin.txt
```diff
@@ -542,6 +542,10 @@ gitsigns.nvim>lua
     gitsigns = true
 <
 
+grug-far.nvim>lua
+    grug_far = true
```
vollowx, line 546: Seems that you've marked this integration as enabled by default, but it's actually not (in [lua/catppuccin/init.lua](https://github.com/catppuccin/nvim/blob/main/lua/catppuccin/init.lua)).

vollowx (changes requested):

doc/catppuccin.txt
mehalter: Ok, cool. I had it as false before and the auto docs generation revised my commit to be true. Is there a way to fix that? I will go ahead and push back the `false` setting

vollowx: Hmm I'm not so familiar with the auto gen tool TBH, but anyway the doc is corrected now :+1:

vollowx (approved):

PR #740, opened 2024-08-02
feat: Add markdown.nvim integration

ValdezFOmar:
Before

image

After

image


ValdezFOmar:
I used the name `render_markdown` since there's already a `markdown` integration, but still tried to clarify in the README that the actual plugin is `markdown.nvim`.

vollowx:
Maybe `markdown_nvim`?

ValdezFOmar:
The plugin [suggests in its README](https://github.com/MeanderingProgrammer/markdown.nvim/tree/main/lua/render-markdown) the use of the name `render-markdown` to avoid collisions with other plugins, the actual [package](https://github.com/MeanderingProgrammer/markdown.nvim/tree/main/lua/render-markdown) uses the same name and the highlight groups and commands use `RenderMarkdown` as a prefix, so maybe is ok?

vollowx:
Makes sense

vollowx (approved):

PR #757, opened 2024-08-10
fix(editor): add missing background for `TabLineFill`

mehalter:
It looks like `TabLineFill` background got removed at some point. This adds it back in


vollowx (approved):
Seems that it's removed when [at] nullchilly was cleaning the outdated `black` colors, glad to see you caught this!

PR #760, opened 2024-08-14
fix(editor): make TabLine more readable

igorlfs:
Hello!

Currently, the `TabLine` foreground is too dark, barely readable with an even darker background:

image

After:

image


vollowx (approved):
LGTM! BTW I wonder if we should modify `TabLineSel` highlight to use the same bg as `Normal` and a `text` fg, in order to present a more modern look， similar to the default colorscheme.

igorlfs:
We could make `TabLine`'s background even darker by using `crust`. Coupled with `TabLineSel = { link = "Normal" }`, here's how it looks:

image

Thoughts?

vollowx:
> make TabLine's background even darker by using `crust`

That make the selected one stand out, looks really nice, and we should also modify `TabLineFill` to use the same `crust` bg

igorlfs:
> That make the selected one stand out, looks really nice, and we should also modify `TabLineFill` to use the same `crust` bg

Hey! I guess it'd be somewhat nice to have the fill with a different color. That's not very common within neovim, but feels more modern IMO. Perhaps @mehalter could have some input, since he (re)introduced `TabLineFill` some weeks ago.

mehalter:
I don't have too much opinion here. If there is a good other color to use then that's probably good if it is more modern. Without it in there at all it was quite distracting so as long as there's a good choice I'm good with it :) plus if I disagree it's easy enough to change locally!

PR #768, opened 2024-09-08
fix(syntax): use `overlay2` for comments

uncenter:
Teeny change from overlay0 to overlay2 for comments to match the [style guide](https://github.com/catppuccin/catppuccin/blob/main/docs/style-guide.md#code-editors).


adriafarres:
Not sure I'm digging this change. It's bright enough that it's harder to discern code from comments, but maybe it's just a matter of getting used to.

PR #776, opened 2024-10-11
fix(editor): more "modern" tabline

igorlfs:
As discussed in https://github.com/catppuccin/nvim/pull/760#pullrequestreview-2239840349

Before:

image

After:

image


vollowx:
That looks quite nice! I guess we need to fix the stylua failure before merging this PR.

igorlfs:
The failure is unrelated, as it's also happening on main. If you push a fix to main, I'll merge it here

vollowx (approved):
Then let's merge this first, and then fix that failure at main

PR #783, opened 2024-10-20
fix(dap-ui): add colors for *NC hlgroups of UI controls

fnune:
`nvim-dap-ui` has highlight groups for all its buttons: `DapUIPlayPause`, `DapUIStepOver`, etc. But it also has `*NC` versions of those highlights. I assume NC stands for non-current or "no cursor".

`nvim-dap-ui` defaults to linking those to their non-`*NC` versions but I assume it happens too early and Catppuccin does not pick that up. This PR simply duplicates the highlight groups to include `*NC` versions.

See https://github.com/rcarriga/nvim-dap-ui/blob/ffa89839f97bad360e78428d5c740fdad9a0ff02/lua/dapui/config/highlights.lua#L75-L96

## Before

Screenshot_20241020_111017
Screenshot_20241020_110955

## After

Screenshot_20241020_110436
Screenshot_20241020_110432

### DIFF
--- a/lua/catppuccin/groups/integrations/dap_ui.lua
+++ b/lua/catppuccin/groups/integrations/dap_ui.lua
@@ -24,13 +24,21 @@ function M.get()
 		DapUIBreakpointsDisabledLine = { fg = C.surface2 },
 
 		DapUIStepOver = { fg = C.blue },
+		DapUIStepOverNC = { link = "DapUIStepOver" },
 		DapUIStepInto = { fg = C.blue },
+		DapUIStepIntoNC = { link = "DapUIStepInto" },
 		DapUIStepBack = { fg = C.blue },
+		DapUIStepBackNC = { link = "DapUIStepBack" },
 		DapUIStepOut = { fg = C.blue },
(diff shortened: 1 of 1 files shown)


lua/catppuccin/groups/integrations/dap_ui.lua
```diff
@@ -24,13 +24,21 @@ function M.get()
 		DapUIBreakpointsDisabledLine = { fg = C.surface2 },
 
 		DapUIStepOver = { fg = C.blue },
+		DapUIStepOverNC = { fg = C.blue },
```
sgoudham, line 27: Hey, thanks for the PR!
Out of curiosity, are we unable to use the `link` function like above with `DapUIFloatBorder` to link the `NC` highlights to the normal ones? 

Something like:

```suggestion
		DapUIStepOverNC = { link = "DapUIStepOver" },
```

I would appreciate if you could let me know if that works or not as I currently can't try it out myself.

fnune: That should work! I pushed a different commit :)

sgoudham (approved):

PR #792, opened 2024-10-27
docs(integrations): reflect that `neotree` is enabled by default

NickyMeuleman:
Document neotree as being an integration that is active by default.
ref: [#c536623](https://github.com/catppuccin/nvim/commit/c536623eac60f8443c93ae4ca0e03b51574b5f50)

### DIFF
--- a/README.md
+++ b/README.md
@@ -951,7 +951,7 @@ mini = {
 <td>
 
 ```lua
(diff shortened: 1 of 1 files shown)


sgoudham (approved):
Nice catch, thanks!

Issue #35, opened 2021-09-20
[Feature] Highlight operators


igorlfs:
The [tokyonight](https://github.com/folke/tokyonight.nvim) theme provides a nice integration with operators:
![image](https://user-images.githubusercontent.com/84649544/133963102-f8c5d2fb-8a81-401c-9141-5118478fee34.png)
as opposed to catpuccino's current implementation:
![image](https://user-images.githubusercontent.com/84649544/133962977-ff80fbcd-bc21-4042-b8fa-c1118779c2fe.png)
(notice the difference between `<<`, `<`, `=`, `++`, `&` and `*`)
Additionally, semicolons and comma are also highlighted, but I think that's a nitpick

Pocco81:
First off, this was intentional, because I didn't know which color to put in order to not make the code look very colorful.

Which color would you like for those? Perhaps a pale pink? Or perhaps something more like the theme you mentioned (cyan)?

Pocco81:
> Oh, I'm sorry if I sounded too demanding...

Oh, sorry if I sounded like I was making you sound demanding... Lol

> A pale pink sounds nice to me!

Nice! I'll test that.

Pocco81:
Enjoy! I ended up choosing cyan :+1:

igorlfs:
> Enjoy! I ended up choosing cyan 

It's pretty good!

Pocco81:
Thank you :)

LuisxSullivaN:
Hey this is awesome

PR #804, opened 2024-11-15
fix(treesitter)!: align token colors with vscode

uncenter:


uncenter:
Fwiw this PR is "ready to merge" but not sure how y'all want to deal with shipping large syntax highlighting changes.

sgoudham:
We'll want to make a bit of an announcement / ship it as a breaking change, along with before/after examples and whatnot.

Let's keep it as a draft for now and revisit later if that's alright :+1:

vollowx:
Really like this for the new HTML look :)

calops:
Hi there, is there a plan to move this forward at some point?

uncenter:
I still have some more adjustments to make for specific languages, and core/staff want to give it a thorough look before sending it through. Hopefully soon though, I'll try to finish up my last changes tonight :)

uncenter:
Quick status update!

I wasn't able to wrap up my work that night two weeks ago, mostly because I've been running into issues with how LSP highlighting ("semantic highlighting") interferes with regular Tree-sitter highlighting and often ends up leaving the appearance worse than before. There are a couple ways around this that I'm looking into (i.e. remapping the LSP captures), but it is taking longer than expected. Sorry for the wait y'all 😅

uncenter:
I think we should merge this as-is (or close to) and avoid other blockers like stuff with LSP highlighting. Once this "major" rework of the syntax highlighting colors goes through, other smaller fixes can also follow.

uncenter:
> I just tested your branch and found that @Property tokens are being assigned the same color as Function tokens, which is visually confusing and doesn't match the vscode equivalent. See attached image (left: vscode, right: nvim).
> compare
> 
> Output from Inspect on **name** token, in case it is useful:
> inspect

Correct, as you can see by the inspect view that coloring is due to the LSP highlighting. I believe if you run the command to temporarily stop the LSP for the session, you will then see the regular/expected property colors? I do plan on correcting the LSP scopes/captures going forward (as mentioned in previous comments) but I don't have the time at the moment to include the required changes for that in this pull request - that being said I will take a closer look this week if this PR isn't merged by then to potentially fix that here.

sbhutch:
Yeah, it is definitely the LSP semantic tokens that invoke the color change, but it appears to be linking the token back to the standard treesitter color for @property tokens. My concern here is that treesitter alone may be smart enough to label variables as properties for some languages and thus color them the same as functions.

I understand if time is an issue here though. I would be happy to help out with LSP semantic token issues for some languages after the merge.

uncenter:
> Yeah, it is definitely the LSP semantic tokens that invoke the color change, but it appears to be linking the token back to the standard treesitter color for @Property tokens. My concern here is that treesitter alone may be smart enough to label variables as properties for some languages and thus color them the same as functions.
> 
> I understand if time is an issue here though. I would be happy to help out with LSP semantic token issues for some languages after the merge.

I appreciate the offer and the concern. I'm now revisiting my thoughts on why exactly I have `@property` set to be blue - trying to think in what case that capture refers to something we do want to be blue? If you could help me investigate that, would be much appreciated. By any chance are you in the Catppuccin Discord so we can communicate in the Neovim dev channel there to avoid polluting this pull request?

sbhutch:
> By any chance are you in the Catppuccin Discord so we can communicate in the Neovim dev channel there to avoid polluting this pull request?

I was not aware of the Discord server, I will catch you in there.

igorlfs:
> I'm quite aware that these highlighting changes might not be received in the best way, especially if people haven't used Catppuccin recently in other editors

Hello,

Apologies if this comes off as petty, but I feel like neovim's original style should have more weight in establishing the style guide. I'm just a regular ([long time](https://github.com/catppuccin/nvim/issues/35)) user, not a member or anything, so I have no idea how these things are defined, but I'd like to give my 2 cents.

Catppuccin started with neovim. As it was ported to other editors, contributors had to make compromises, either because there was no style guide back then, and they had to fill in the gaps, or because the tool lacked highlighting as good as treesitter's. Of course, further discrepancies might have been introduced as time passed, due to a lack of coordination. But I feel like neovim's implementation should still be **the** reference, as not only it was the first one, but also due to neovim's highly flexible highlighting capabilities.

To me, this change sounds like: "VS Code is more popular, therefore other tools should adopt its style".

> Thankfully, if people prefer the highlights from before this PR, they can make use of our custom overrides system to get the previous highlights back:

By the way, if reconsidering this change is completely off question, could we have an option for using the original style? So people don't have to override over 30 highlight groups on their own.

DanWlker:
> It can be contributed. I don't use Neovim enough to care on the finer details, maybe somone like @sbhutch can jump in here.

In that case can this pr be reverted? Seems like a half baked fixed with huge implications, if the semantic token implementation is not coming soon. Still, thank you for the effort of syncing the colors.

>It's not about which highlighter is used, it's just which is a more accurate living representation of what the org wants to do with the palette. VS Code has been receiving constant tweaks and refinements that were never carried over to Neovim, hence this large pull request to sync the two. Tree-sitter also isn't much better of a reference because every editor does Tree-sitter differently.

Doesn't treesitter provide a more "precise" way of highlighting when compared to vscode? What defines accurate here? is there a discussion page or reference about the choices for syntax highlighting? Genuinely wanting to look into it.

gabrielscaranello:
Hello everyone, I'd like to express my feelings about this change.  
I usually agree with most of the changes, and I think they were made for a good reason. But for me, the colors, at least in TypeScript, don't make sense.

| Before | After |
| --- | --- |
| before | after |

uncenter:
> Hello everyone, I'd like to express my feelings about this change. I usually agree with most of the changes, and I think they were made for a good reason. But for me, the colors, at least in TypeScript, don't make sense.
> Before 	After
> before 	after

The only inaccuracy I see here besides the semantic tokens issue is `Promise` being `mauve` from `@type.builtin`, which isn't solvable. The semantic tokens issue I will look into again now.

sgoudham:
@uncenter and I had a look into this a bit more, and it looks like we missed out on updating the semantic tokens which is a layer on top of treesitter.  I've created a new issue (#900) to track the update, and we won't be releasing v2.0.0 until that's fixed. Apologies!

With regards to the other comments, I'll be creating an v2.0.0 discussion post for people to leave their comments instead of this PR. Here's a small summarised list of my thoughts atm:

1. The introduction of these changes is not because VSCode is more popular, but rather VSCode was the first port we started with when evolving/creating the first iteration of the style guide from the tokens within Neovim.
2. Colours are, ultimately, subjective. Myself and winston (previous core maintainer) agreed that some highlights in the neovim port felt quite overbearing and wanted to trim those back a little, which may come across as "regressions" - To clarify, I'm all for improving syntax highlighting in a meaningful way, but I'm not interested throwing all of our accent colours into all possible treesitter captures.
3. The style guide 100% needs a revamp/many improvements and @unseen-ninja and I will hopefully be tackling that when time permits into a more detailed and meaningful guide with examples, including code editors. The `catppuccin/catppuccin` repository is definitely the place to raise issues if you wish to give feedback on the syntax highlighting.
4. This might seem a little mean, but I'm not interested in reverting this PR or introducing a new layer of options for "legacy" syntax highlighting. I'd encourage the usage of the custom highlights API via the snippet in my comment above if you prefer the colours before this PR.

RayGuo-ergou:
> I don't see how it is acceptable to completely change the well established colorscheme. One of the most user hostile things I have ever seen

This is one of the most arrogant and entitled statement i see towards an open source project.

sgoudham:
Any feedback with regards to these changes/v2.0.0 should be posted under https://github.com/catppuccin/nvim/discussions/903.

PR #808, opened 2024-11-25
feat(gitsign): improve gitsign inline add and delete colors

AndreasNasman:
This PR aims to improve the colors of changed text when running `:Gitsigns preview_hunk` with `transparent_background = false` (see the screenshot below).

For the dark themes, the cursor color seems to "conflict" with the diff colors, making it hard to track the cursor if focusing a hunk and, e.g., copying some text.

For `GitSignsAddInline`, it took the color of https://github.com/catppuccin/nvim/blob/faf15ab0201b564b6368ffa47b56feefc92ce3f4/lua/catppuccin/groups/syntax.lua#L73 and doubled the amount from `18` to `36`. I did the same with `GitSignsDeleteInline` and https://github.com/catppuccin/nvim/blob/faf15ab0201b564b6368ffa47b56feefc92ce3f4/lua/catppuccin/groups/syntax.lua#L75

# Latte

## Before

image

## After

image

# Frappe

## Before

image

## After

image

# Macchiato

## Before

image

## After

image

# Mocha

## Before

image

## After

image

### DIFF
--- a/lua/catppuccin/groups/integrations/gitsigns.lua
+++ b/lua/catppuccin/groups/integrations/gitsigns.lua
@@ -46,6 +46,9 @@ function M.get()
 
 			GitSignsAddPreview = O.transparent_background and { fg = C.green, bg = C.none } or { link = "DiffAdd" },
 			GitSignsDeletePreview = O.transparent_background and { fg = C.red, bg = C.none } or { link = "DiffDelete" },
+
+			GitSignsAddInline = { bg = U.darken(C.green, 0.36, C.base) },
+			GitSignsDeleteInline = { bg = U.darken(C.red, 0.36, C.base) },
 		}
 	end
 end


AndreasNasman:
I'm unsure if I should add some `O.transparent_background` checks. Since https://github.com/catppuccin/nvim/blob/faf15ab0201b564b6368ffa47b56feefc92ce3f4/lua/catppuccin/groups/integrations/gitsigns.lua#L8 already does an outer check for the truthiness of `O.transparent_background`, I don't follow why the inner checks like https://github.com/catppuccin/nvim/blob/faf15ab0201b564b6368ffa47b56feefc92ce3f4/lua/catppuccin/groups/integrations/gitsigns.lua#L47 are needed.

vollowx:
> I'm unsure if I should add some `O.transparent_background` checks. Since
> 
> https://github.com/catppuccin/nvim/blob/faf15ab0201b564b6368ffa47b56feefc92ce3f4/lua/catppuccin/groups/integrations/gitsigns.lua#L8
> 
> already does an outer check for the truthiness of `O.transparent_background`, I don't follow why the inner checks like
> https://github.com/catppuccin/nvim/blob/faf15ab0201b564b6368ffa47b56feefc92ce3f4/lua/catppuccin/groups/integrations/gitsigns.lua#L47
> 
> are needed.

That makes sense, we could fix this in another PR.

vollowx (approved):

PR #809, opened 2024-12-10
feat(blink-cmp): improve styling support

milkias17:
add hl groups for BlinkCmpKind, BlinkCmpMenu, BlinkCmpMenuBorder, BlinkCmpDocBorder, BlinkCmpSignatureHelpActiveParameter

- This was the previous styling for blink.cmp, it didn't have the correct background color for the completion menu, no border styling and no styling for showing the current signaturehelp item.
Screenshot_20241210_163953

- This is how it looks after the new hl group additions:
image

- All styling colors were taken from previous nvim-cmp styling and the catppuccin vscode theme.

### DIFF
--- a/lua/catppuccin/groups/integrations/blink_cmp.lua
+++ b/lua/catppuccin/groups/integrations/blink_cmp.lua
@@ -9,6 +9,7 @@ function M.get()
 		BlinkCmpMenuBorder = { fg = C.blue },
 		BlinkCmpDocBorder = { fg = C.blue },
 		BlinkCmpSignatureHelpActiveParameter = { fg = C.mauve },
+		BlinkCmpSignatureHelpBorder = { fg = C.blue },
 
 		BlinkCmpLabelMatch = { fg = C.text, style = { "bold" } },
 		BlinkCmpKindText = { fg = C.green },
--- a/lua/catppuccin/groups/integrations/blink_cmp.lua
+++ b/lua/catppuccin/groups/integrations/blink_cmp.lua
@@ -4,6 +4,11 @@ function M.get()
 	return {
 		BlinkCmpLabel = { fg = C.overlay2 },
 		BlinkCmpLabelDeprecated = { fg = C.overlay0, style = { "strikethrough" } },
+		BlinkCmpKind = { fg = C.blue },
+		BlinkCmpMenu = { fg = C.text },
+		BlinkCmpMenuBorder = { fg = C.blue },
+		BlinkCmpDocBorder = { fg = C.blue },
+		BlinkCmpSignatureHelpActiveParameter = { fg = C.mauve },
 
 		BlinkCmpLabelMatch = { fg = C.text, style = { "bold" } },
 		BlinkCmpKindText = { fg = C.green },


lua/catppuccin/groups/integrations/blink_cmp.lua
```diff
@@ -4,6 +4,11 @@ function M.get()
 	return {
 		BlinkCmpLabel = { fg = C.overlay2 },
 		BlinkCmpLabelDeprecated = { fg = C.overlay0, style = { "strikethrough" } },
+		BlinkCmpKind = { fg = C.blue },
+		BlinkCmpMenu = { fg = C.text },
+		BlinkCmpMenuBorder = { fg = C.blue },
+		BlinkCmpDocBorder = { fg = C.blue },
+		BlinkCmpSignatureHelpActiveParameter = { fg = C.mauve },
```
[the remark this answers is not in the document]
milkias17: My bad, have added it now

vollowx:
By default we assume that the user uses the un-bordered UI and apply the colors for that, if you still want to support the bordered UI, please add a option named `bordered` and seperate the highlight groups.

milkias17:
@vollowx Perhaps I'm missing something, and correct me if I'm wrong here.

BlinkCmpMenuBorder highlight group only takes effect when someone enables the bordered UI, otherwise there's no difference in the UI for the user. And you said we apply the colors for that, however I'm not changing an existing highlight group, the BlinkCmpMenuBorder highlight group doesn't exist in the current blink.cmp integration.

So I'm having trouble understanding your objection, would love to be corrected if I'm missing something tho.

lua/catppuccin/groups/integrations/blink_cmp.lua
```diff
@@ -4,6 +4,12 @@ function M.get()
 	return {
 		BlinkCmpLabel = { fg = C.overlay2 },
 		BlinkCmpLabelDeprecated = { fg = C.overlay0, style = { "strikethrough" } },
+		BlinkCmpKind = { fg = C.blue },
+		BlinkCmpMenu = { fg = C.text },
```
vollowx, line 8: I was looking at the screenshots and found that the background has been reset, and I think it is this line that changed the background. You probably need to explicitly declare the background color in this case.

vollowx (changes requested):

lua/catppuccin/groups/integrations/blink_cmp.lua
milkias17: I'm not sure what exactly you mean by "the background has been reset", the current blink.cmp integration doesn't set the background color of the completion menu(the screenshot at the top), what I did was change the background of the menu to be reflective of the catppuccin colors.

The appearance of the menu(with the catppuccin background) is the exact same ui that's in place for nvim-cmp

vollowx: By default, according to [this line](https://github.com/Saghen/blink.cmp/blob/main/lua/blink/cmp/highlights.lua#L32) in blink.cmp, the `BlinkCmpMenu` is set to be the same as `Pmenu`, which has a background color of darkened `surface0` according to [this line](https://github.com/catppuccin/nvim/blob/4965db2d6155c25db4e8417465fc2703fdf4c2b7/lua/catppuccin/groups/editor.lua#L44). After defining the `BlinkCmpMenu` manually without setting the background color, the background becomes `base`, which doesn't look good when using the un-bordered UI.

PR #812, opened 2024-12-22
fix: match fzf-lua highlights to telescope

lopi-py:
The background color doesn't match the border, making it inconsistent. The title is now consistent with telescope too.

Before
image

After
image

### DIFF
--- a/lua/catppuccin/groups/integrations/fzf.lua
+++ b/lua/catppuccin/groups/integrations/fzf.lua
@@ -2,11 +2,14 @@ local M = {}
 
 function M.get()
 	return {
-		FzfLuaNormal = { link = "NormalFloat" },
+		-- FzfLuaNormal = { link = "NormalFloat" }, Respect fzf-lua's default float bg
 		FzfLuaBorder = { link = "FloatBorder" },
-		FzfLuaTitle = { link = "FloatTitle" },
+		FzfLuaTitle = { link = "FloatBorder" },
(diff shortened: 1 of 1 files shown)


sgoudham (approved):

lua/catppuccin/groups/integrations/fzf.lua
```diff
@@ -2,9 +2,9 @@ local M = {}
 
 function M.get()
 	return {
-		FzfLuaNormal = { link = "NormalFloat" },
+		-- FzfLuaNormal = { link = "NormalFloat" },
```
sgoudham, line 5: Is there a reason this should stay commented out? 🤔

lopi-py: I saw that telescope integration [did the same](https://github.com/catppuccin/nvim/blob/5d9582dbbe37e465f7f93ca6e3dd60cd20eea44f/lua/catppuccin/groups/integrations/telescope.lua#L48). lmk if it needs a little explanation or if I just remove it

sgoudham: Ah didn't realise Telescope had the same. I think a little comment/explanation like the one on Telescope would be good!

lopi-py:
I have changed the prompt and matching colors a bit to be more close to telescope too

Before
image
After
image

sgoudham (approved):

PR #825, opened 2025-01-29
feat: add Avante integration

joshmedeski:
Adding support for [avante](https://github.com/yetone/avante.nvim). I'm open to changing what colors are used on the titles, I'm not sure if we have a convention for that.

## Screenshots

Sidebar colors:
SCR-20250129-laaf

Inline text
SCR-20250129-labm

If rounded is turned off
SCR-20250129-kxud

### DIFF
--- a/lua/catppuccin/groups/integrations/avante.lua
+++ b/lua/catppuccin/groups/integrations/avante.lua
@@ -1,0 +1,57 @@
+local M = {}
+
+function M.get()
+	local rounded = O.integrations.avante.windows_sidebar_header_rounded
+	return {
+		-- titles
+		AvanteTitle = rounded and { bg = C.lavender, fg = C.base } or { fg = C.lavender },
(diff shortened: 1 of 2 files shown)


exsesx:
Great, looking forward to this being merged! 

I'd love to see support for suggestions as well. Check out the `AvanteToBeDeleted` highlight and similar cases.

image

joshmedeski:
Grouped the colors, improved the conflict colors, and added some missing color groups, thanks @exsesx for the suggestion

lua/catppuccin/init.lua
```diff
@@ -39,6 +39,7 @@ local M = {
 		default_integrations = true,
 		integrations = {
 			alpha = true,
+			avante = false,
```
vollowx, line 42: You don't need to explicitly declare that this integration is disabled by default.

vollowx (changes requested):

PR #826, opened 2025-02-09
fix(gitsigns): respect transparent option for gitsigns

heddxh:
If I am not misunderstand, https://github.com/catppuccin/nvim/pull/461#issuecomment-1518543190 means there is a `transparent` option for gitsigns integrations. So this PR really respect this option and edit the doc.


vollowx:
Could you please also give the before/after screenshot?

heddxh:
Sorry for the late reply. And here we goes.

Before:

Before

After: 

After

All with `transparent_background = true` and:

```      
integrations = {
  gitsigns = {
    enabled = true,
    transparent = false,
  },
},
```

emxxjnm:
Hi @heddxh and @vollowx, I believe `gitsigns.transparent` should align with the `transparent_background` option by default instead of being separate. Additionally, we need to update the types.lua and README.

heddxh:
If `gitsigns.transparent` and `transparent_background` is both true, then word_diff will not actually work, just like my screenshot above. So I think maybe keep `gitsigns.transparent = false` as default is better?

PR #833, opened 2025-02-20
feat(gitgutter): add LineNr highlights

DOD-101:
Add [LineNr highlights](https://github.com/airblade/vim-gitgutter?tab=readme-ov-file#line-number-highlights) for the gitgutter integration: 

image


vollowx:
Thanks for your contribution!

PR #836, opened 2025-02-24
fix(octo): incorrect highlight backgrounds

### LINKED ISSUE #835: Octo: Some highlights have incorrect background color
### Description

There are some elements with a wrong-looking background in the Octo PR view. I'm not sure if that's intentional but they look wrong and original Octo (with integration disabled) doesn't have any difference with the main background.

Original colors:
Image

Catppuccin integration colors:
Image

### Neovim version

```markdown
NVIM v0.10.4
Build type: Release
LuaJIT 2.1.1713773202
```

### Terminal and multiplexer

Ghostty 1.0.1, tmux 3.5a

Latest commit (4bb938bbba41d306db18bf0eb0633a5f28fd7ba0)

### Steps to reproduce

1. Enable `octo` integration (example using `lazy`):
```lua
return {
  'catppuccin/nvim',
  name = 'catppuccin',
  priority = 1000,
  opts = {
    integrations = {
      octo = true,
    },
  },
  config = function(_, opts)
    require('catppuccin').setup(opts)
    vim.cmd.colorscheme 'catppuccin'
  end,
}
```
2. Open any PR in Octo: `Octo pr edit 123>

### Expected behavior

The backgrounds of all elements (bubbles, PR state, author and reviewers) should be the same as the rest of the buffer, just like with original Octo colors.

### Actual behavior

Some elements have darker backgrounds.

TimShilov:
Fixed octo integration highlight backgrounds.

The only thing that I'm not fully happy with is "Created by" bubble which, I believe, should have a different background but it seems to be using the same `OctoBubble` as the `Reviewers` section so I couldn't find a way to make it look better.
I might give it another try later, but it looks way better than in both original and catppuccin versions already.

Original Octo colors (for the reference):
Screenshot 2025-02-24 at 00 57 34

Catppuccin integration colors (BEFORE):
Screenshot 2025-02-24 at 00 57 52

Catppuccin integration colors (AFTER):
Screenshot 2025-02-24 at 00 58 08

Fixes #835.


sgoudham (approved):
Thanks!

PR #847, opened 2025-04-03
fix(telescope): nvchad and transparency

Lalit64:
Hi, Here is a small fix to basically make it so that if the `telescope.style = "nvchad"` and `transparent_background` = true, then telescope will be opaque and non transparent. I think that if the user explicitly specifies that `telescope.style = "nvchad"` the user would want the telescope window to be opaque.

### DIFF
--- a/lua/catppuccin/groups/integrations/telescope.lua
+++ b/lua/catppuccin/groups/integrations/telescope.lua
@@ -13,15 +13,21 @@ function M.get()
 			},
 			TelescopePromptBorder = {
 				fg = O.transparent_background and C.blue or C.surface0,
-				bg = O.transparent_background and C.none or C.surface0,
+				bg = O.transparent_background and O.integrations.telescope.style == "nvchad" and C.surface0
+					or C.none
+					or C.surface0,
 			},
 			TelescopePromptNormal = {
 				fg = C.text,
-				bg = O.transparent_background and C.none or C.surface0,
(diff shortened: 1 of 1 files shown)


vollowx:
Thanks for your contributions, please also attach a before/after screenshot so that we can review this more easily.

Lalit64:
With config:
```lua
opts = {
  flavour = "mocha",
  integrations = {
    telescope = {
      enabled = true,
      style = "nvchad",
    },
  },
}
```

Before:
CleanShot 2025-04-03 at 19 52 05@2x

After:
CleanShot 2025-04-03 at 19 51 11@2x

comfysage:
i think there should probably be a `opts.integrations.telescope.transparent` option that should default to the global `transparent_background` setting. this way users can still maintain the old behavior.

Lalit64:
@comfysage I think that if the user wants nvchad enabled he would want it to be opaque? If the user doesn't have nvchad enabled and transparency enabled it is still normal and not opaque.

uncenter:
I agree with @comfysage here; it makes the most sense to me to add an option to allow users who have enabled `transparent_background` globally to disable transparency specifically for this integration to attain the behavior you refer to.

Lalit64:
@uncenter `style = "nvchad"` is the option to disable transparency as of this pr. If it is not defined then it is by default transparent. Why would the user do `style = "nvchad"` and then want to enable transparency in telescope?

Lalit64:
Hey can someone check this out?

vollowx:
Please check #877

PR #850, opened 2025-05-04
feat(signify): add vim-signify integration

andreihh:
Use the same highlights as for gitgutter / gitsigns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1586,6 +1586,20 @@ sandwich = false
 </tr>
 <!-- vim-sandwich -->
 
+<!-- vim-signify -->
+</tr>
+<tr>
+<td> <a href="https://github.com/mhinz/vim-signify">vim-signify</a> </td>
+<td>
+
+```lua
+signify = true
+```
+
+</td>
+</tr>
+<!-- vim-signify -->
+
 <!-- vim-sneak -->
 </tr>
 <tr>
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1586,6 +1586,20 @@ sandwich = false
...
 <!-- vim-sandwich -->
 
+<!-- vim-signify -->
+</tr>
+<tr>
+<td> <a href="https://github.com/mhinz/vim-signify">vim-signify</a> </td>
+<td>
+
+```lua
+signify = true
```
vollowx, line 1596: We only enable general and very popular integrations by default, so please keep this disable by default

lua/catppuccin/init.lua
```diff
@@ -110,6 +110,7 @@ local M = {
 				enabled = true,
 				indentscope_color = "overlay2",
 			},
+			signify = true,
```
vollowx, line 113: Also remove this line

vollowx (changes requested):
Thank you for your contribution ❤️ 

Here're some places you need to modify before merging

README.md
andreihh: Done!

lua/catppuccin/init.lua
andreihh: Done!

andreihh:
Should be ready now.

vollowx (approved):

PR #857, opened 2025-05-11
fix(cursor): update cursor color to rosewater / crust

andreihh:
Follow the style guide at
https://github.com/catppuccin/catppuccin/blob/main/docs/style-guide.md.

Ports for popular terminals follow this style, so nvim, being a primarily terminal app, should strive the same.


andreihh:
Note: the style guide recommends using `base` instead of `crust` for `latte`, but I'm not sure how to properly check which flavor is enabled at runtime. However, `base` and `crust` are very similar for `latte`, so maybe this is a good-enough approximation?

vollowx (approved):
LGTM ❤️

PR #861, opened 2025-05-25
fix: transparent markdown headings bg

### LINKED ISSUE #845: Headings background color not set in render-markdown.nvim integration on transparent mode on latte variant.
### Description

Hello! Using `transparent_background = true` in the options, the background colors of headings in markdown files is just raw black with the Latte variant. This is not the color I get in non transparent mode and is very hard to read in my opinion.

Image

I found that the file https://github.com/catppuccin/nvim/blob/main/lua/catppuccin/groups/integrations/render_markdown.lua sets the base to `nil` if the background is transparent, which might be the cause of this. Thank you in advance for your help!

### Neovim version

```markdown
NVIM v0.10.2
Build type: Release
LuaJIT 2.1.1713773202
```

### Terminal and multiplexer

kitty 0.39.1

catppuccin main branch, commit 5b5e3ae

### Steps to reproduce

1. Write a markdown `test.md` file with some heading of any level, such as `## Level 2 heading`
2. nvim -u repro.lua test.md

### Expected behavior

The headings have a readable background, same as non transparent-mpde if possible.

### Actual behavior

They have a black background, very hard to read.

### Repro

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
end

-- install plugins
local plugins = {
  { "catppuccin/nvim", config = function() require("catppuccin").setup({ transparent_background = true }) end },
  "MeanderingProgrammer/render-markdown.nvim",
}
  root = root .. "/plugins",
})

vim.cmd.colorscheme("catppuccin-latte")
-- add anything else here
```

comfysage:
use crust as a darkening background for markdown headings with transparent background. previously would fall back to `#000000` (`nil`) which does not work well for latte.

fixes #845 

before-image
after-image

BEGIN_COMMIT_OVERRIDE
fix(markview): incorrect heading bg when using transparent bg
fix(render-markdown): incorrect heading bg when using transparent bg
END_COMMIT_OVERRIDE

### DIFF
--- a/lua/catppuccin/groups/integrations/markview.lua
+++ b/lua/catppuccin/groups/integrations/markview.lua
@@ -32,7 +32,7 @@ function M.get()
 	}
 
 	local syntax = require("catppuccin.groups.syntax").get()
-	local base = not O.transparent_background and C.base or nil
+	local base = not O.transparent_background and C.base or C.crust
 
 	for i = 0, 7 do
 		local color = rainbow[i] or syntax["rainbow" .. i].fg
--- a/lua/catppuccin/groups/integrations/markview.lua
+++ b/lua/catppuccin/groups/integrations/markview.lua
@@ -32,11 +32,10 @@ function M.get()
 	}
 
 	local syntax = require("catppuccin.groups.syntax").get()
-	local base = not O.transparent_background and C.base or C.crust
 
 	for i = 0, 7 do
 		local color = rainbow[i] or syntax["rainbow" .. i].fg
-		local bg = U.darken(color, darkening_percentage, base)
+		local bg = U.darken(color, darkening_percentage, C.base)
 		groups["MarkviewPalette" .. i] = { fg = color, bg = bg }
 		groups["MarkviewPalette" .. i .. "Fg"] = { fg = color }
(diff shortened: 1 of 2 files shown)


comfysage:
image

lua/catppuccin/groups/integrations/markview.lua
```diff
@@ -32,11 +32,10 @@ function M.get()
 	}
 
 	local syntax = require("catppuccin.groups.syntax").get()
-	local base = not O.transparent_background and C.base or nil
 
 	for i = 0, 7 do
 		local color = rainbow[i] or syntax["rainbow" .. i].fg
-		local bg = U.darken(color, darkening_percentage, base)
+		local bg = U.darken(color, darkening_percentage, C.base)
```
sgoudham, line 38: Screenshots look good, just one general comment here:

It's interesting that this logic seemed to be passing `nil` into the `U.darken()` function when there was no transparent background. I'd like to make sure the behaviour is still correct without a transparent background? @comfysage 

I briefly skimmed the utils.lua code but need to spend some more time understanding this codebase.

comfysage: passing nil to the darken function just made it default to pure black

PR #867, opened 2025-06-06
feat(blink-cmp): add `style` option

comfysage:
adds a `style` to configure whether to use bordered or solid borders. 

can in a future pr be used to add atom-like pmenu styling.

Closes #809

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,9 @@ beacon = false
 <td>
 
 ```lua
-blink_cmp = false
+blink_cmp = {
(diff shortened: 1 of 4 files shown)


vollowx:
@milkias17 Can you have a look on this PR, since this should have fixed the issue you were trying to solve in #809 ?

dragonlobster:
bordered style should explicitly have a different color for `BlinkCmpMenu`, because the border style, the link to `Pmenu` doesn't look right:

image

PR #871, opened 2025-06-16
feat: automatic integration detection

### LINKED ISSUE #838: feat: automatic integrations
For now every needed plugin needs to be specified in `integrations`. It would be easier to setup if the plugins could be detected and enabled automatically.

For reference, tokyonight.nvim supports such feature by `{ plugins = { auto = true } }`.

0xmayro:
closes #838 

detects installed plugins and creates a matching integrations table.
currently only plugins installed with [lazy.nvim](https://github.com/folke/lazy.nvim) are detected.

I need to check how to populate the config's integrations table with the generated one and provide an option for it.

I would like to get some comments on the location in the code base.
any other feedback is also appreciated!

### DIFF
--- a/lua/catppuccin/init.lua
+++ b/lua/catppuccin/init.lua
@@ -197,7 +197,6 @@ function M.setup(user_conf)
 	end
 
 	M.options = vim.tbl_deep_extend("keep", user_conf, M.default_options)
-
 	M.options.highlight_overrides.all = user_conf.custom_highlights or M.options.highlight_overrides.all
 
 	-- Get cached hash
--- a/lua/catppuccin/init.lua
+++ b/lua/catppuccin/init.lua
@@ -37,6 +37,7 @@ local M = {
 			operators = {},
 		},
 		default_integrations = true,
+		auto_integrations = false,
 		integrations = {
 			alpha = true,
 			blink_cmp = true,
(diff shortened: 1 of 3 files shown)


lua/catppuccin/init.lua
```diff
@@ -173,9 +174,22 @@ function M.setup(user_conf)
...
+	if user_conf.auto_integrations == true then
+		if user_conf.integrations ~= nil then
+			vim.tbl_deep_extend(
+				"keep",
+				user_conf.integrations,
+				require("catppuccin.lib.detect_integrations").create_integrations_table()
+			)
+		else
+			user_conf.integrations = require("catppuccin.lib.detect_integrations").create_integrations_table()
+		end
```
comfysage, line 186:
```suggestion
		user_conf.integrations = vim.tbl_deep_extend(
			"force",
			require("catppuccin.lib.detect_integrations").create_integrations_table(),
			user_conf.integrations or {}
		)
```

lua/catppuccin/types.lua
```diff
@@ -33,6 +33,8 @@
 ---@field styles CtpStyles?
 -- Should default integrations be used.
 ---@field default_integrations boolean?
+---@field auto_integrations boolean?
+-- Should detect integrations automatically?
```
comfysage, line 37:
```suggestion
-- Should detect integrations automatically
```

comfysage (changes requested):

lua/catppuccin/init.lua
```diff
@@ -187,6 +197,7 @@ function M.setup(user_conf)
 	end
 
 	M.options = vim.tbl_deep_extend("keep", user_conf, M.default_options)
+
```
comfysage, line 200:
```suggestion
```

comfysage (changes requested):
everything looks great 👍 

thnx for the pr its hugely appreciated

just some final touches :)

vollowx (approved):
LGTM!

Would be better if you can merge the process of building a list up from `lazy.nvim` into `M.create_integrations_table`, so that we only need to travel through the list once.

PR #881, opened 2025-07-25
feat: plugin to integration mappings table generator

0xmayro:
this pr adds a script to generate the integration_mappings table used in the auto detection of installed plugins.

additionally it adds a line to each file in the `lua/catppuccin/groups/integrations/` directory
to note the plugin name as it will appear in the table received from lazy.nvim  

note:
the integrations/plugins are not alphabetically organized in the generated table.
I have no experience in github actions so couldn't test it, please let me know if I need to change anything to allow it.


scripts/generate_integration_mappings_table.lua
```diff
@@ -0,0 +1,54 @@
...
+-- create and write the table to target_file
+local target_file = io.open(target_path, "w+")
+assert(target_file ~= nil, "a target file must be specified")
+
+target_file:write "local M = {\n"
+for _, line in ipairs(lines) do
+	target_file:write(line)
+end
+target_file:write "}\n"
+target_file:write "\nreturn M"
```
vollowx, line 52: Need a \n at the end

PR #877, opened 2025-07-22
feat: make bg of `FloatBorder` same as `NormalFloat`

vollowx:
feat(telescope): add `nvchad_outlined` style

dup #847 #874 

@Lalit64 @Atifchy Please check if this PR works for you.

- `1`: doesn't look different before/after
- `2`: doesn't look different whether `transparent_background` is `true` or `false`
- solid: `transparent_background = false`
- transparent: `transparent_background = true`

Single border windows:

| * | before | after |
| --- | --- | --- |
| transparent (`1`) |  |  |
| solid | Sway_20250722_20h29m40s | Sway_20250722_20h29m28s |

`telescope.nvim`:

| * | before | after |
| --- | --- | --- |
| transparent `classic` (`1`) |  |  |
| solid `classic` (`1`) |  |  |
| transparent `nvchad` | Sway_20250722_20h38m04s | Sway_20250722_20h38m17s |
| solid `nvchad` (`1`) |  |  |
| `nvchad_outlined` (`2`) |  | Sway_20250722_20h35m11s |

PR #885, opened 2025-07-28
feat!: add global floating window options

### LINKED ISSUE #856: transparent_background should not make NormalFloat transparent

I've recently upgraded to the latest version. Had to mess a lot with transparency, which was working as expected before.

With the upgrade, setting the `transparent_background = true` (an option I did not have before) every backgrounds are now transparent. Making it very hard to see elements such as floats and any other UI element which do not have borders.

Almost all NormalFloat elements do not seem to have borders so I believe this should be the default behavior.

transparent_background should not override NormalFloat elements transparency. These element background should respect regular theme.

**Describe alternatives you've considered**

I was able to mitigate the problem by doing:
```
        return {
          NormalFloat = { bg = colors.base, fg = colors.text },
        }
      end,
```

comfysage:
fixes #856

based on the work started in #877

putting all the floating window options, that were previously in integration opts, globally so its clear what integrations should provide in customization:
```lua
-- config:
{
  float = {
    transparent = false, -- whether floating windows should be transparent
    solid = false, -- whether to use nvchad (solid) style or classic style floats
  },
}
```

in the current implementation i would get the following results for `transparent_background = true`:
trans bg + trans float + classic float = outlined
trans bg + opague float + solid float = nvchad
trans bg + trans float + solid float = nvchad outlined

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ require("catppuccin").setup({
         dark = "mocha",
     },
     transparent_background = false, -- disables setting the background color.
+    float = {
+        transparent = false, -- enables transparency on floating windows
+        solid = false, -- use nvchad styling for floating windows
+    },
     show_end_of_buffer = false, -- shows the '~' characters after the end of buffers
     term_colors = false, -- sets terminal colors (e.g. `g:terminal_color_0`)
     dim_inactive = {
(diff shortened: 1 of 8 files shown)


Kayzels:
Is there a reason this was reverted? I've been explicitly overriding the highlight groups to get this effect of solid floats, but transparent background. And then I saw the commit today, but it was reverted around an hour and a half later? Did it break something?

sgoudham:
Sorry, should have explained in the description!

We meant to release v1.11.0 with all of our minor/patch changes before this breaking one. It'll be merged again soon in #892

PR #886, opened 2025-07-28
fix: make sure to enable auto detected integrations with table configurations

mehalter:
The new auto integration detection is great! I noticed in some cases it doesn't enable the integration unless I manually put `enabled = true` in the integrations table. I dug into this and it appears it's missing the portion that sets `enabled = true` for table based integration configuration.

The example I ran into this with was setting up custom configuration for `snacks = { picker_style = "nvchad" }` in my integrations table and the snacks integration is not properly enabled when I start the editor. It simply needs `enabled = true` enabled. This will also be the case for something like `colorful_winsep` or `navic` where the default configuration has `enabled = false`


vollowx (approved):
LGTM

PR #892, opened 2025-07-31
feat!: add global floating window options

comfysage:
reopen #885

PR #893, opened 2025-07-31
fix(snacks): fix snacks picker input field color

AtifChy:
Screenshot 2025-07-31 111017

fixed snacks picker input field bg color, now it has the same color as rest of the window.


comfysage:
this is fixed in #892

PR #909, opened 2025-08-09
ci: auto generate integrations mapping table

0xmayro:
I accidentally messed up the [old PR](https://github.com/catppuccin/nvim/pull/881) trying to fix a different one.

this should be the same.


0xmayro:
@vollowx @sgoudham , again I'm so sorry for all the mess.

vollowx (approved):
LGTM 👍

scripts/generate_integration_mappings_table.lua
```diff
@@ -0,0 +1,51 @@
...
+	local mappings = {}
+	for filename, _ in vim.fs.dir(integrations_path) do
+		filename = vim.fn.fnamemodify(filename, ":r")
+
+		local ok, mod = pcall(require, "catppuccin.groups.integrations." .. filename)
+
+		if ok then
+			---@type string
+			local plugin_name = mod.url
+			plugin_name = plugin_name:match ".*/(.-)/*$" or plugin_name -- extract the repo from the url. else use the url.
```
comfysage, line 21: can this pattern be replaced with `([^/]+)$`?

scripts/generate_integration_mappings_table.lua
```diff
@@ -0,0 +1,51 @@
...
+			mappings[plugin_name] = filename
```
comfysage, line 22:
```suggestion
			do
				---@type string
				local plugin_name = mod.url
				if not plugin_name then
					return
				end
				plugin_name = plugin_name:match ".*/(.-)/*$" or plugin_name -- extract the repo from the url. else use the url.
				mappings[plugin_name] = filename
			end
```

comfysage (changes requested):
looks great :) thnx for all the hard work

just some small changes id like to see

scripts/generate_integration_mappings_table.lua
0xmayro: this is for making sure mod.url is correctly set, right?
instead of `return` should it print an error with the file name?

comfysage:
summary of required changes (as described in collaboration with @sgoudham ):
- move logic out to utility function (e.g., getting the name from url) and write tests.
- compare `M.url` with the name of the repo according to the plugin-spec's urls from `vim.pack`, `lazy.nvim`, `pckr`.
- merge #910 into this pr.

if you need help with any of these id love to help write some of the logic/tests

0xmayro:
about this, should the utility function be defined in a new file? or can the script return it?
i.e. 
```lua
-- script.lua
local M = {}
function M.get_name(url)
	return url:match("([^/]+)$")
end
-- rest of script as defined in PR...

return M
```

comfysage (approved):

sgoudham (approved):
Trusting that it's good to merge from @comfysage

PR #910, opened 2025-08-09
feat!: support `vim.pack` and `pckr.nvim`

0xmayro:
adds support for auto detection of plugins installed through [pckr.nvim](https://github.com/lewis6991/pckr.nvim) and [vim..pack](https://github.com/neovim/neovim/pull/34009)


0xmayro:
merged into #909

Issue #919, opened 2025-08-18
Bufferline integration not working as expected


comfysage:
@vollowx @mehalter @comfysage  this commit broke bufferline color

<img width="529" height="78" alt="image" src="https://github.com/user-attachments/assets/bb58b0e1-af14-4d17-8e6c-6a57392408d2" />

_Originally posted by @AtifChy in https://github.com/catppuccin/nvim/pull/918#issuecomment-3198021155_

could you provide some additional information and repro? would love to help solve this issue

AtifChy:
> The default colorscheme in LazyVim is Tokyo night and not catppuccin. The screenshot shared is not this colorscheme.

You switch them using <leader>uC

mehalter:
This is probably less related to `auto_integrations` and just the setup of the bufferline integration itself. It appears if you have the `bufferline` integration enabled then the get function will just return an empty table rather than the integration function.

Here is a better minimal configuration that doesn't rely on all of a distribution:

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system "curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua")()

require("lazy.minit").repro {
  spec = {
    {
      "catppuccin/nvim",
      name = "catppuccin",
      opts = {
        integrations = {
          bufferline = true,
        },
      },
      config = function(_, opts)
        require("catppuccin").setup(opts)
        vim.cmd.colorscheme "catppuccin"
      end,
    },
    {
      "akinsho/bufferline.nvim",
      version = "*",
      dependencies = "nvim-tree/nvim-web-devicons",
      opts = function(_, opts) opts.highlights = require("catppuccin.groups.integrations.bufferline").get() end,
    },
  },
}
```

mehalter:
My guess is this is probably present in multiple different plugins such as `feline` which also changes the semantics of what `get()` does. Most of the integations when you run `get()` you get a table of highlight groups and their definitions, but these plugins that require special setup change the semantics to `get()` returning something plugin specific.

comfysage:
> My guess is this is probably present in multiple different plugins such as `feline` which also changes the semantics of what `get()` does. Most of the integations when you run `get()` you get a table of highlight groups and their definitions, but these plugins that require special setup change the semantics to `get()` returning something plugin specific.

yes this is probably the actual feature. it doesnt help that lazyvim abstracts plugin configuration so much that people aren't able to provide proper repro's.

[...]

PR #920, opened 2025-08-18
fix(integrations)!: rename `get()` to `get_*()` on abnormal integrations

comfysage:
- **fix(lib/mapper): allow for integrations without `get()` field**
  

- **fix(integrations/feline): rename `get()` to `get_statusline()`**
  

- **fix(integrations/bufferline): rename `get()` to `get_theme()`**

theres an issue where some integrations should never be able to get enabled through the config options: if they do it will error because the returned value will be assumed to be a highlights table. this will obviously cause issues when the returned value is not meant to be passed to `vim.api.nvim_set_hl()`, like with feline and bufferline.

this issue is occuring because the `auto_integrations` enables these special integrations if the plugin is detected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -579,7 +579,7 @@ local ctp_feline = require('catppuccin.groups.integrations.feline')
 ctp_feline.setup()
 
 require("feline").setup({
-    components = ctp_feline.get(),
+    components = ctp_feline.get_statusline(),
 })
 ```
 
@@ -668,7 +668,7 @@ vim.api.nvim_create_autocmd("ColorScheme", {
         package.loaded["feline"] = nil
         package.loaded["catppuccin.groups.integrations.feline"] = nil
         require("feline").setup {
-            components = require("catppuccin.groups.integrations.feline").get(),
+            components = require("catppuccin.groups.integrations.feline").get_statusline(),
         }
     end,
 })
--- a/README.md
+++ b/README.md
@@ -400,7 +400,7 @@ use "akinsho/bufferline.nvim" {
   after = "catppuccin",
   config = function()
     require("bufferline").setup {
-      highlights = require("catppuccin.groups.integrations.bufferline").get()
+      highlights = require("catppuccin.groups.integrations.bufferline").get_theme()
     }
   end
 }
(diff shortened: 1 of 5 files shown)


vollowx (approved):
I think this is a good solution, yet I'd also recommend to divide integrations into core/plugins/plugins_special in the file system, so it would be like

```
- groups
  editor.lua
  markdown.lua # I personally think that md and semantic tokens should be in core rather than plugins
  semantic_tokens.lua
  ...
  - plugins
    drop_bar.lua
    ...
  - special
    bufferline.lua
    ...
```

comfysage:
> Does this fix #919 ?

yes it does seem to be related. i dont know how lazyvim sets this colorscheme up but the bufferline integration currently still works - its just that `auto_integrations` doesnt work for bufferline or feline.

comfysage:
> I think this is a good solution, yet I'd also recommend to divide integrations into core/plugins/plugins_special in the file system, so it would be like

i would rather put the special integrations in `/lib/plugins/`. it doesnt make sense to put them in the `groups` dir since they do not contain highlight groups. this means other integrations can just stay where they are.
i do agree that markdown and semantic tokens should probably on the same level in the filesystem as `treesitter` and `editor`.

sgoudham (approved):
Trusting that it works, thanks 👍

vollowx:
> > I think this is a good solution, yet I'd also recommend to divide integrations into core/plugins/plugins_special in the file system, so it would be like
> 
> 
> 
> i would rather put the special integrations in `/lib/plugins/`. it doesnt make sense to put them in the `groups` dir since they do not contain highlight groups. this means other integrations can just stay where they are.
> 
> i do agree that markdown and semantic tokens should probably on the same level in the filesystem as `treesitter` and `editor`.

The `lib` and `utils` kind of feel similar, for example I think `hsluv.lua` and `hash.lua` would also fit in utils, and it doesn't really make much sense to divide these things into two folders. Meanwhile if we put special plug-in integrations in the `lib` folder, it might be even more confusing. If we are going to change it and it's not in the `groups` folder, `lua/catppuccin/special` might be a fine path, as it has nothing to do with the rest of the folders semantically.

For the markdown and semantic tokens, I would definitely suggest to move them out of `integrations`, since the integrations table generator is also affected by them.

PR #921, opened 2025-08-18
ci: merge patch workflows

comfysage:


.github/workflows/patch.yml
```diff
@@ -0,0 +1,49 @@
+name: Patch
+on:
```
sgoudham, line 2: Easy way to re-run if we want to for whatever reason

```suggestion
on:
	workflow_dispatch:
```

.github/workflows/patch.yml
```diff
@@ -1,7 +1,6 @@
 name: Patch
 on:
   pull_request:
-  push:
     branches:
       - main
```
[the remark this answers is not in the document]
sgoudham: That should be fine. That code is saying run on all pull requests to the main branch.

.github/workflows/patch.yml
```diff
@@ -0,0 +1,52 @@
+name: Patch
+on:
+  workflow_dispatch:
+  push:
```
sgoudham: `on: push` shouldn't have been added back in @comfysage

comfysage: uhm whoopsie, sorry

.github/workflows/patch.yml
```diff
@@ -0,0 +1,50 @@
+name: Patch
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+
+jobs:
+  integration-table-gen:
+    needs: [docs]
```
sgoudham: This is an insane nitpick, sorry in advance lol, but can you move `integration-table-gen` underneath `docs` to also make the order similar to the `needs`

PR #925, opened 2025-09-07
:zap: Make neotree transparency seperated from global configs

mhkarimi1383:
🎉 First off, thanks for taking the time to contribute! 🎉

Here are some guidelines:
- Format code using [stylua](https://github.com/johnnymorganz/stylua).
- Create a topic branch on your fork for your specific PR.
- Use [conventionalcommits.org's](https://www.conventionalcommits.org/en/v1.0.0/)
  rules for explicit and meaningful commit messages.
- If it's your first time contributing to a project, then read
  [About pull requests](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests)
  on Github's docs.

Here are some tips:
- Use `vim.g.catppuccin_debug = true` to get live config re-loading

### DIFF
--- a/README.md
+++ b/README.md
@@ -1013,7 +1013,10 @@ mini = {
 <td>
 
 ```lua
-neotree = true
+neotree = {
+    enabled = true,
+    transparent = false,
+}
 ```
 
 </td>
--- a/doc/catppuccin.txt
+++ b/doc/catppuccin.txt
@@ -677,7 +677,10 @@ mini.nvim>lua
 <
 
 neo-tree.nvim>lua
-    neotree = true
+    neotree = {
+        enabled = true,
+        transparent = false,
+    }
 <
 
 neogit>lua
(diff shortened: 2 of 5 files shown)


mhkarimi1383:
Also why the background color of highlighted item is not different?

this makes searches harder and `follow_current_file` useless

comfysage:
hi with [#892](<https://github.com/catppuccin/nvim/pull/892>) we've moved away from integration-specific transparency options and instead rely on global options.
if you want to override the transparency for certain highlights you can use [custom highlights](https://github.com/catppuccin/nvim#overwriting-highlight-groups) in your config.

in this case you could use:
```lua
require("catppuccin").setup {
        return {
            NeoTreeWinSeparator = { fg = colors.surface1, bg = colors.none },
        }
    end
}
```

PR #928, opened 2025-09-14
fix(integrations): update neogit url

trimclain:
This fixes the typos in #909.
If I understand it correctly, the wrong mapping in [integration_mappings.lua](https://github.com/catppuccin/nvim/blob/main/lua/catppuccin/utils/integration_mappings.lua#L41) should be fixed by the CI. Let me know if I should correct them myself.


comfysage:
i just pushed it myself for now since we'll be refactoring the patch workflows with #921

sgoudham (approved):

PR #948, opened 2025-10-17
fix(treesitter): use text for html heading content

comfysage:
RE: https://github.com/catppuccin/nvim/discussions/903#discussioncomment-14656351


comfysage:
> Any reason why we can't link `@markup.heading` to `@markup`? Does that unintended side effects?

i cant find any place where it is used, so i dont know if there are any side effects.

PR #949, opened 2025-10-17
refactor!: remove vim support

### LINKED ISSUE #932: move support to nvim
exclusively support nvim, meaning that support for vim will be dropped. those who prefer to use vim can use the catppuccin vim port instead.

the vim portion of this port has become unmaintained and currently only bloats the already way too massive codebase of this project.
the vim and neovim project has grown so far apart over time that maintaining both has become unmanagable and will always mean compromising on either one of the projects.

see: https://github.com/catppuccin/nvim/issues/940#issuecomment-3393263233

comfysage:
fixes #932 

steps:
- [x] remove vim lib
- [x] remove vim checks
- [x] replace viml with lua
- [x] update documentation 
- [x] add `vim` branch (2fc720c6)

### DIFF
--- a/doc/catppuccin.txt
+++ b/doc/catppuccin.txt
@@ -50,6 +50,20 @@ with Neovim 0.12’s `vim.pack`
 <
 
 
+VIM SUPPORT ~
+
+Support for vim is provided through the `vim` branch.
+
+>vim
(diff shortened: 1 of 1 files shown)


comfysage:
> This is a straight up removal, not deprecation right?

youre right. i used 'deprecate' since support is still available under `catppuccin/vim` but ill change the phrasing.

PR #953, opened 2025-11-14
feat: add blink.indent integration

HeitorAugustoLN:
Adds an integration for the [blink.indent](https://github.com/saghen/blink.indent) plugin


lua/catppuccin/init.lua
```diff
@@ -64,6 +64,7 @@ local M = {
 		integrations = {
 			alpha = true,
 			blink_cmp = { enabled = true, style = "bordered" },
+			blink_indent = { enabled = true },
```
comfysage:
```suggestion
			blink_indent = true,
```

PR #954, opened 2025-11-15
fix(editor): editor items in visual selection issue

mao-yining:
The color of quickfix line is still not obvious. So I suggest to use mauve instead.


comfysage:
this change does not follow the catppuccin [style guide](https://github.com/catppuccin/catppuccin/blob/main/docs/style-guide.md) and can cause legibility problems.
there's also some unrelated changes with Todo in here that shouldnt be included.

PR #959, opened 2025-11-26
Fix: string index error causes failed load in kitty

mathematicalninja:
```lua
local red_green_string = hex:sub(1, 5)
local blue_value = tonumber(hex:sub(6, 7), 16)
```

would read the first 5 digits as red-green and the last digit as blue

then
```lua 
ans[accent] = string.format("%s%.2x", red_green_string, blue_value)
```
would format the 6th digit to be 2 characters wide.

So the hex_to_rgb would parse a 7-digit hex code and crash

### DIFF
--- a/lua/catppuccin/palettes/init.lua
+++ b/lua/catppuccin/palettes/init.lua
@@ -18,8 +18,8 @@ function M.get_palette(flavour)
 	--]]
 	if O.kitty then
 		for accent, hex in pairs(ans) do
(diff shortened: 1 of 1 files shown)


comfysage:
im sorry if this is a stupid question but isn't the (1, 5) including the hash sign at the start of the hex code, so yes it would be a 7 char hex code: `#XXXXXX`?

PR #966, opened 2026-01-08
fix: Change FloatShadow from fg to bg color

Old-Farmer:
Hello catppuccin/nvim maintainers. I have found a bug that `set winborder=shadow` doesn't work in this colorscheme and I make a fix by changing the FloatShadow hl group.


comfysage (approved):
thnx for the contribution <3

PR #972, opened 2026-03-03
feat(artio): add artio.nvim integration

comfysage:
adds integration for artio fuzzy finder:
https://github.com/comfysage/artio.nvim

image


vollowx (approved):
Perhaps should not be added in the default integrations? Beside that remember to update #976

PR #973, opened 2026-03-03
ci: fix patch workflow

comfysage:
from what i can read this should fix the repeated issues with auto-commit.

https://docs.github.com/en/actions/reference/workflows-and-actions/variables
https://github.com/stefanzweifel/git-auto-commit-action/tree/v7-next


vollowx (approved):
LGTM ♥️

Issue #980, opened 2026-03-17
[Suggestion] A better implementation for hiding EOB characters


lutalli:
By default, this theme hides the `~` characters at the end of a buffer and it seems to do so by setting its color to the same as the background.

This method doesn't work well with transparent background, as there the `~` characters will still be more or less visible.

Maybe a cleaner workaround would be to set the EOB filling character to a space. In Lua:

```lua
vim.opt.fillchars = { eob = " " }
```

This has been working smoothly for me.

comfysage:
yeah i noticed this too when i took over maintenance of the theme. its one of the smaller changes id like to get into a larger pr at some point.

i would just change the fg color as overlay0 and leave it up to users to hide it if they want.

thnx for pointing this out <3

PR #982, opened 2026-03-21
fix!: remove `O.show_end_of_buffer`

comfysage:
removing this option since its behavior is dependent on a non-transparent background.

if users want to keep this, they should use:
```lua
vim.opt.fillchars:append({ eob = " " })
```

fix https://github.com/catppuccin/nvim/issues/980


vollowx (approved):
Nice to have this breaking change ❤️ 

Perhaps we should also mention this change in the 2.0.0 release note in case someone doesn’t know how to hide eob chars with vim.o.list

PR #1007, opened 2026-05-20
feat(hl/editor): add Dimmed hl

comfysage:
i dont know if we want to link related highlight groups to this new `Dimmed` group or keep the original `fg = C.overlay1`.
some hls this could work for:
- `NeoTreeDimText`
- `Conceal`
- `NeoTestIndent` (which should probably be using the same as `IblIndent` (suface0)
- `NeoTestExpandMarker`
- `SagaCount`
- `DropBarIconUISeparator`

https://github.com/neovim/neovim/pull/39505

### DIFF
--- a/doc/catppuccin.txt
+++ b/doc/catppuccin.txt
@@ -1,4 +1,5 @@
 *catppuccin.txt*                              Soothing pastel theme for NeoVim
+                                 For nvim >= 0.8.0    Last change: 2026 May 20
 
 ==============================================================================
 Table of Contents                               *catppuccin-table-of-contents*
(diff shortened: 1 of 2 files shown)


vollowx:
> if we want to link related highlight groups to this new Dimmed group

Considering the purpose of `Dimmed`, I think doing so would be nice for
- `NeoTreeDimText`
- `NeoTestExpandMarker`
- `SagaCount`
- `DropBarIconUISeparator`

(from what you've listed).

However, these hls might get linked within the plugins, and this isn't a very blocking problem, so it should be fine to merge this MR before doing all that.

vollowx (approved):

PR #1011, opened 2026-05-31
feat(editor): add highlights for PmenuKind and PmenuKindSel

andreihh:
The default links to Pmenu and PmenuSel, which makes the kind text or symbol difficult to distinguish from the actual item.

Configured PmenuKind and PmenuKindSel to match Pmenu and PmenuSel in `bg` and `style`, but set `fg` to `blue` (same as for `blink.cmp` and `nvim-cmp`).

Before (kind symbols have same color as item):
catppuccin-pmenukind-before

After (kind symbols are blue to distinguish them from items):
catppuccin-pmenukind-after

### DIFF
--- a/lua/catppuccin/groups/editor.lua
+++ b/lua/catppuccin/groups/editor.lua
@@ -71,6 +71,11 @@ function M.get()
 		PmenuMatchSel = { style = { "bold" } }, -- Popup menu: matching text in selected item; is combined with |hl-PmenuMatch| and |hl-PmenuSel|.
 		PmenuSbar = { bg = C.surface0 }, -- Popup menu: scrollbar.
 		PmenuThumb = { bg = C.overlay0 }, -- Popup menu: Thumb of the scrollbar.
+		PmenuKind = {
+			bg = (O.transparent_background and vim.o.pumblend == 0) and C.none or C.mantle,
+			fg = C.blue,
+		}, -- Popup menu: item kind. Background and style must match `Pmenu`.
+		PmenuKindSel = { bg = C.surface0, fg = C.blue, style = { "bold" } }, -- Popup menu: selected item kind. Background and style must match `PmenuSel`.
 		PmenuExtra = { fg = C.overlay0 }, -- Popup menu: normal item extra text.
 		PmenuExtraSel = {
 			bg = C.surface0,


lua/catppuccin/groups/editor.lua
```diff
@@ -70,6 +70,11 @@ function M.get()
 		PmenuMatchSel = { style = { "bold" } }, -- Popup menu: matching text in selected item; is combined with |hl-PmenuMatch| and |hl-PmenuSel|.
 		PmenuSbar = { bg = C.surface0 }, -- Popup menu: scrollbar.
 		PmenuThumb = { bg = C.overlay0 }, -- Popup menu: Thumb of the scrollbar.
+		PmenuKind = {
+			bg = (O.transparent_background and vim.o.pumblend == 0) and C.none or C.mantle,
```
andreihh, line 75: Is it necessary to repeat the `bg` / `style` here, or will it be inherited from `Pmenu` / `PmenuSel`?

andreihh: Nvm, I checked just now and yes, this is necessary.

andreihh: This is how it looks without explicit `bg` (note the inconsistent kind column color):

Image

This is how it looks with explicit `bg`:

Image

vollowx:
Hi, can you attach a before/after screenshot so that we can better review this? I cannot use Neovim at this moment

andreihh:
Hi! I updated the PR description with before/after screenshots. Hope this helps clarify the change.

comfysage:
whats the reasoning behind the color blue for the item kind?

andreihh:
It's what the non-specific [`BlinkCmpKind`](https://github.com/catppuccin/nvim/blob/c7c692a0ad3080710893abbae100171819a3e4be/lua/catppuccin/groups/integrations/blink_cmp.lua#L11) and [`CmpItemKind`](https://github.com/catppuccin/nvim/blob/c7c692a0ad3080710893abbae100171819a3e4be/lua/catppuccin/groups/integrations/cmp.lua#L9) link to, so I went with that for consistency.

But I don't have a preference, I'm happy with anything (e.g., Peach for a stronger contrast) as long as as there's a clear visual distinction from the item text.

comfysage:
i feel like a neutral color (e.g. subtextX, text) would be a better fit here ~especially to avoid the "rainbowpuke" we have been trying to avoid since v2.

andreihh:
Here's an example for subtext1:
catppuccin-pmenukind-subtext1

IMO there's still no distinction at all between the kind and the text.

If you prefer to avoid rainbowpuke, then I think blue is a better fit? It's more muted / weaker contrast to the surrounding text, and maintains consistency with the highlights for the most popular completion plugins.

vollowx:
Yeah I would agree to use blue, as it is a pretty general and not opinionated color.

And there probably wouldn't be a lot of users noticing this since colourful "kind" is configured by many. They can customise this anyway.

andreihh:
Is there anything that's still blocking this?

andreihh:
Thank you!

PR #1019, opened 2026-07-19
fix!: remove `default_integrations`

### LINKED ISSUE #1018: docs: arbitrary default integrations
the integrations that are enabled by default seem kind of arbitrary ~im a little ensure (when adding new integrations) whether they should be added to the defaults table.

i propose we either come up with a better selection process for these integrations or disable all integrations by default.

fix #1017

comfysage:
fix #1018 

additional changes:
- feat: enable `auto_integrations` by default
- fix(integrations): options can be nil
- fix(integrations): use `O` over `require("catppuccin").options`
- fix(colorful_winsep): color option can be invalid
- fix(gitsigns): option logic
- fix(indent_blankline): do not index table with nil
- fix(mini): color option can be invalid
- fix(navic): do not use `CtpColor` for color fields
- fix(snacks): color option can be invalid
- fix(telescope): make integration options a boolean this was a table even though it didnt have any options

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,8 +168,7 @@ require("catppuccin").setup({
     },
     color_overrides = {},
     custom_highlights = {},
-    default_integrations = true,
-    auto_integrations = false,
+    auto_integrations = true,
     integrations = {
         cmp = true,
         gitsigns = true,
@@ -295,15 +294,7 @@ require("catppuccin").setup({
 })
 ```
 
-Some integrations are enabled by default, you can control this behaviour with `default_integrations` option.
-
-```lua
-require("catppuccin").setup({
-    default_integrations = false,
-})
-```
-
-If you use [lazy.nvim](https://github.com/folke/lazy.nvim) as your package manager, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations.
+If you use [lazy.nvim](https://github.com/folke/lazy.nvim) as your package manager, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations (this option is enabled by default).
 
 ```lua
 require("catppuccin").setup({
--- a/README.md
+++ b/README.md
@@ -294,7 +294,11 @@ require("catppuccin").setup({
 })
 ```
 
-If you use `vim.pack` for managing your plugins, or if you [lazy.nvim](https://github.com/folke/lazy.nvim) or [pckr.nvim](https://github.com/lewis6991/pckr.nvim) as plugin managers, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations (this option is enabled by default).
+If you use `vim.pack`, [lazy.nvim](https://github.com/folke/lazy.nvim) or
+[pckr.nvim](https://github.com/lewis6991/pckr.nvim) to manage your plugins, you
+can use the `auto_integrations` option to let catppuccin automatically detect
+installed plugins and enable their respective integrations (this option is
+enabled by default).
 
 ```lua
 require("catppuccin").setup({
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ require("catppuccin").setup({
 })
 ```
 
-If you use [lazy.nvim](https://github.com/folke/lazy.nvim) as your package manager, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations (this option is enabled by default).
+If you use `vim.pack` for managing your plugins, or if you [lazy.nvim](https://github.com/folke/lazy.nvim) or [pckr.nvim](https://github.com/lewis6991/pckr.nvim) as plugin managers, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations (this option is enabled by default).
 
 ```lua
 require("catppuccin").setup({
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -295,15 +294,7 @@ require("catppuccin").setup({
...
-Some integrations are enabled by default, you can control this behaviour with `default_integrations` option.
-
-```lua
-require("catppuccin").setup({
-    default_integrations = false,
-})
-```
-
-If you use [lazy.nvim](https://github.com/folke/lazy.nvim) as your package manager, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations.
+If you use [lazy.nvim](https://github.com/folke/lazy.nvim) as your package manager, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations (this option is enabled by default).
```
vollowx: Better mention all `vim.pack`, lazy.nvim and pckr.nvim?

lua/catppuccin/init.lua
```diff
@@ -55,66 +55,8 @@ local M = {
 				background = true,
 			},
 		},
-		default_integrations = true,
-		auto_integrations = false,
-		integrations = {
-			alpha = true,
-			blink_cmp = { enabled = true, style = "bordered" },
```
vollowx, line 62: I guess we should keep the `style` field or set the default in the blink-cmp file? Otherwise we are keeping default colors but not such options

lua/catppuccin/init.lua
```diff
@@ -55,66 +55,8 @@ local M = {
...
-			neogit = true,
-			neotree = true,
-			nvimtree = true,
-			ufo = true,
-			rainbow_delimiters = true,
-			render_markdown = true,
-			telescope = { enabled = true },
-			treesitter_context = true,
-			barbecue = {
-				dim_dirname = true,
```
vollowx, line 82: Also here, removing them directly might cause runtime errors or other behaviors

vollowx (changes requested):
That's a huge overhaul!

About the default options, I think it'd be fine to keep them, just removing the `enabled` field. It would be more complicated and harder to maintain if the default values are set all over the place.

comfysage:
i agree about keeping the default options inside the table but i think the integrations should still handle nil values better. handling nil values is also necessary to prevent lsp warnings.

README.md
```diff
@@ -294,7 +294,7 @@ require("catppuccin").setup({
 })
 ```
 
-If you use [lazy.nvim](https://github.com/folke/lazy.nvim) as your package manager, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations (this option is enabled by default).
+If you use `vim.pack` for managing your plugins, or if you [lazy.nvim](https://github.com/folke/lazy.nvim) or [pckr.nvim](https://github.com/lewis6991/pckr.nvim) as plugin managers, you can use the `auto_integrations` option to let catppuccin automatically detect installed plugins and enable their respective integrations (this option is enabled by default).
```
vollowx:
```
... or if you [lazy-nvim] ...
             ^
             use
``` 

or perhaps

```
If you use `vim.pack`, [lazy.nvim](https://github.com/folke/lazy.nvim) or [pckr.nvim](https://github.com/lewis6991/pckr.nvim) to manage your packages
```

vollowx:
> i agree about keeping the default options inside the table but i think the integrations should still handle nil values better. handling nil values is also necessary to prevent lsp warnings.

Do you mean LSP warnings inside this project or at the user config? Inside this project we can safely assume that directly accessing such field will not result in `nil`, so perhaps modifying the type declarations helps. I cannot verify either of them right now so I'm not sure about the user side one.

comfysage:
> Do you mean LSP warnings inside this project or at the user config? Inside this project we can safely assume that directly accessing such field will not result in `nil`, so perhaps modifying the type declarations helps. I cannot verify either of them so I'm not sure about the user side one.

the config options should stay nillable to avoid diagnostic warnings in user configs, so we cant change the type definitions. checking for nil is often really simple and in my mind required to avoid bugs. additionally, it is necessary to check for nil so that the lsp can check if the actual type of the option is used correctly in the integration.

sorry for the giant wall of text

vollowx (approved):

vollowx:
> the config options should stay nillable to avoid diagnostic warnings in user configs, so we cant change the type definitions. checking for nil is often really simple and in my mind required to avoid bugs. additionally, it is necessary to check for nil so that the lsp can check if the actual type of the option is used correctly in the integration.
> 
> sorry for the giant wall of text

That's reasonable I think.

Originally I thought that falling back on `nil` checks might let developers ignore some initializing problems, and not falling back makes sure that you get an error when providing neither a default value nor a user config value.

But since the LSP is not recognizing the types from Neovim table merging, it might be better to handle `nil`s for better LSP type checking, as you said.

When do you think shall we merge this? After adding `nil` checking or leave it in another PR?

comfysage:
i already added nil checking in this pr:)

vollowx:
I mean for those that are not colors, like those in barbecue and blink cmp. But it'd be fine not to do so now since the default values are present.

I think we can merge now?

comfysage:
> I mean for those that are not colors, like those in barbecue and blink cmp. But it'd be fine not to do so now since the default values are present.

these dont really need type checking since theyre either bools (and nil is falsy) or enum strings (in which case theyre always compared to a string anyways). so for all of these cases the lsp is totally able to parse the types.

i can still add explicit nil checking in a separate pr if youd like?

vollowx:
I think we don't need to since the LSP works on this and default values are passed, so it'd all good

PR #1022, opened 2026-07-21
fix(integrations): auto_integrations not created if not set by user

mbaklor:
After #1019 enabled `auto_integrations` by default, the actual code path to check `auto_integrations` was still using the user supplied setup options. This meant that unless explicitly set to `true` by the user, the `auto_integrations` are effectively still false.
Moved the check to after merging with the default options.

I checked this on my system and it works but I don't have a ton of custom options set so I'm not entirely sure this is the exact correct call.

### DIFF
--- a/lua/catppuccin/init.lua
+++ b/lua/catppuccin/init.lua
@@ -149,16 +149,16 @@ function M.setup(user_conf)
 	-- Parsing user config
 	user_conf = user_conf or {}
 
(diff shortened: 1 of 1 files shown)


comfysage:
thnx for the helpful debugging<3
it helped me figure out a simple way to fix the issue.
im closing this now in favor of #1023

PR #1023, opened 2026-07-22
fix: check if `auto_integrations` was explicitly disabled

comfysage:
`auto_integrations` is enabled by default now which means that simply checking if the user enabled the option does not work. Instead, we should check if `auto_integrations == false`.


vollowx (approved):

vollowx:
Though this patch would work, IMO it'd be better to merge the user config first - doing it like this essentially bypasses the one in the default config and is weird semantically.

Merge: options = user_config & default_config
Check: options.auto_integrations?
Yes?: options = options & auto_config (since there are nothing that overlaps between default and auto ones, their order does not matter)
No?: continue

(Left ones have higher priority)

(Sorry for the format, I'm writing on the phone)

vollowx (approved):

